PR #62, opened 2025-12-22
feat(sdk): added interfaces

ittaysw:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/62)

### DIFF
--- a/sdk/src/interfaces.ts
+++ b/sdk/src/interfaces.ts
@@ -34,11 +34,11 @@ export type ViewingKey = BigNumberish;
 
 export type StarknetAddress = BigNumberish;
 
-export const witnessBrand: unique symbol = Symbol("witness");
-
-export type Witness = {
-  readonly [witnessBrand]: true;
-};
+/** Base class for witness objects. Provides nominal typing. */
+export abstract class Witness {
+  /** @internal */
+  protected readonly _witnessMarker = undefined;
+}
 
 export type WitnessSerde<B extends BlobT = string> = Serde<Witness, B>;
 
@@ -83,11 +83,11 @@ export type PrivateTransfersConfig = {
   pool: StarknetAddress;
 };
 
-export const channelBrand: unique symbol = Symbol("channel");
-
-export type Channel = {
-  readonly [channelBrand]: true;
-};
+/** Base class for channel objects. Provides nominal typing. */
+export abstract class Channel {
+  /** @internal */
+  protected readonly _channelMarker = undefined;
+}
 
 export type ChannelSerde<B extends BlobT = string> = Serde<Channel, B>;
 
--- a/sdk/src/interfaces.ts
+++ b/sdk/src/interfaces.ts
@@ -70,16 +70,6 @@ export type PrivateTransfersConfig = {
   pool: StarknetAddress;
 };
 
-export type PrivacyState = {
-  timestamp: BlockIdentifier;
-
-  // token -> notes
-  notes: Map<StarknetAddress, Note[]>;
-
-  // recipient -> channel
-  recipients: Map<StarknetAddress, Channel>;
-};
-
 export interface PrivateRecipient {
   address: StarknetAddress;
   context: Channel | Note.Id; // note id is for semi-transparent (preprepared) notes.
@@ -204,19 +194,21 @@ export interface PrivateTransfers {
   build(): PrivateTransfersBuilder;
 
   /**
-   * Discover unspent notes per token and recipient states
+   * Discover unspent notes per token
    *
-   * @recipient discover a specific recipient's state
    */
-  discover({
-    since,
-    known,
-    recipient,
-  }: {
-    since?: BlockIdentifier;
-    known?: PrivacyState;
-    recipient?: StarknetAddress;
-  }): Promise<PrivacyState>;
+  discoverNotes(params: { since?: BlockIdentifier; known?: Map<StarknetAddress, Note[]> }): {
+    timestamp: BlockIdentifier;
+    notes: Map<StarknetAddress, Note[]>;
+  };
+
+  /**
+   * Discover channels for one or more recipients
+   */
+  discoverChannels(...recipients: (StarknetAddress | PrivateRecipient)[]): {
+    timestamp: BlockIdentifier;
+    channels: Map<StarknetAddress, Channel>;
+  };
 }
 
 // ============ Builder Types ============
@@ -355,22 +347,33 @@ export interface ProofProviderInterface {
 
 export interface DiscoveryProviderInterface {
   /**
-   * Discover currently owned notes and state of past recipients.
+   * Discover unspent notes per token
+   *
    */
-  discover(
+  discoverNotes(
     address: StarknetAddress,
     viewingKey: ViewingKey,
-    {
-      since,
-      known,
-      recipient,
-    }: {
-      since?: BlockIdentifier;
-      known?: PrivacyState;
-      recipient?: StarknetAddress;
-    }
-  ): Promise<PrivacyState>;
+    params: { since?: BlockIdentifier; known?: Map<StarknetAddress, Note[]> }
+  ): {
+    timestamp: BlockIdentifier;
+    notes: Map<StarknetAddress, Note[]>;
+  };
 
+  /**
+   * Discover channels for one or more recipients
+   */
+  discoverChannels(
+    address: StarknetAddress,
+    viewingKey: ViewingKey,
+    ...recipients: (StarknetAddress | PrivateRecipient)[]
+  ): {
+    timestamp: BlockIdentifier;
+    channels: Map<StarknetAddress, Channel>;
+  };
+
+  /**
+   * Viewing key of the compliance council for registration
+   */
   globalViewingKey(): ViewingKey;
 }
 
--- a/sdk/src/interfaces.ts
+++ b/sdk/src/interfaces.ts
@@ -18,15 +18,6 @@ export const open: Open = { __marker: "open" };
 export type Self = { readonly __marker: "self" };
 export const self: Self = { __marker: "self" };
 
-type BlobT = string | Uint8Array;
-
-export type Blob<B extends BlobT> = B & { readonly __blob: unique symbol };
-
-export type Serde<T, B extends BlobT> = {
-  encode(from: T): Blob<B>;
-  decode(from: Blob<B>): T;
-};
-
 /**
  * Union that allows both the concrete Account class as well as the lighter AccountInterface.
  */
@@ -34,13 +25,9 @@ export type ViewingKey = BigNumberish;
 
 export type StarknetAddress = BigNumberish;
 
-/** Base class for witness objects. Provides nominal typing. */
-export abstract class Witness {
-  /** @internal */
-  protected readonly _witnessMarker = undefined;
-}
-
-export type WitnessSerde<B extends BlobT = string> = Serde<Witness, B>;
+// Import and re-export Witness class from internal.ts
+import { Witness, Channel } from "./internal.js";
+export { Witness, Channel };
 
 export type Note = {
   readonly id: Note.Id;
@@ -82,14 +69,6 @@ export type PrivateTransfersConfig = {
   discoveryProvider: DiscoveryProviderInterface;
   pool: StarknetAddress;
 };
-
-/** Base class for channel objects. Provides nominal typing. */
-export abstract class Channel {
-  /** @internal */
-  protected readonly _channelMarker = undefined;
-}
-
-export type ChannelSerde<B extends BlobT = string> = Serde<Channel, B>;
 
 export type PrivacyState = {
   timestamp: BlockIdentifier;
@@ -394,3 +373,14 @@ export interface DiscoveryProviderInterface {
 
   globalViewingKey(): ViewingKey;
 }
+
+type BlobT = string | Uint8Array;
+
+export type Blob<B extends BlobT> = B & { readonly __blob: unique symbol };
+
+export type Serde<T, B extends BlobT> = {
+  encode(from: T): Blob<B>;
+  decode(from: Blob<B>): T;
+};
+export type WitnessSerde<B extends BlobT = string> = Serde<Witness, B>;
+export type ChannelSerde<B extends BlobT = string> = Serde<Channel, B>;
--- a/sdk/src/interfaces.ts
+++ b/sdk/src/interfaces.ts
@@ -30,7 +30,7 @@ import { Witness, Channel } from "./internal.js";
 export { Witness, Channel };
 
 export type Note = {
-  readonly id: Note.Id;
+  readonly id: NoteId;
   readonly amount: Amount;
   readonly created?: BlockNumber; // required to know maturity (10 blocks)
   readonly witness: Witness;
@@ -38,10 +38,8 @@ export type Note = {
   readonly sender: StarknetAddress;
 };
 
-export namespace Note {
-  /** Unique identifier for a note, used for semi-transparent (preprepared) notes */
-  export type Id = BigNumberish;
-}
+/** Unique identifier for a note, used for semi-transparent (preprepared) notes */
+export type NoteId = BigNumberish;
 
 export type Proof = {
   readonly data: Uint8Array;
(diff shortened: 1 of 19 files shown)


sdk/src/interfaces.ts
```diff
@@ -0,0 +1,307 @@
...
+  readonly call: Call;
+  readonly proof: Proof;
+};
+
+export type PrivateInvocationResult = {
+  readonly invocationData: CallAndProof;
+  readonly remainder?: Note;
+};
+
+export type PrivateTransfersConfig = {
```
sean-starkware, line 59: Is this not relevant anymore? it is not used. If so, lets delete, if it should be used somehow, can you explain how?

sdk/src/interfaces.ts
```diff
@@ -0,0 +1,307 @@
...
+  ): Promise<PrivateInvocationResult>;
+
+  /**
+   * Withdraw tokens from the privacy pool
+   *
+   * @param amount if not provided, the total amount of the notes is used. if provided and is lower than the total amount of notes, the remainder note is returned. In that case, the channel must be provided
+   *
+   * v1: the inputs size is 1 and the amount must match the output amount.
+   */
+  withdraw(
```
sean-starkware: I think that selfChannel should be required, because by the time that this will be actually executed, there can be a new note created for me with this token, since i don't have control of who sends me money.

So I can think I'm giving the full amount, but I'm actually not, and I can miss the selfChannel. 

Since selfChannel should anyway always be available to the user of the interface, I think it's best to provide one every time.

sdk/src/interfaces.ts
```diff
@@ -0,0 +1,307 @@
...
+    selfChannel?: Channel
+  ): Promise<PrivateInvocationResult>;
+
+  /**
+   * transfer tokens from the privacy pool to a single recipient. 
+   * @param amount if not provided, the total amount of the notes is used. if provided and is lower than the total amount of notes, the remainder note is returned
+
+   * Note: one can send notes with total that exceeds the amount ot transfer, that way a single note is returned as remainder that replaces them.
+   */
+  transfer(
```
sean-starkware: same here for selfChannel to be required

sdk/src/interfaces.ts
```diff
@@ -0,0 +1,307 @@
...
+   *       calldata: [...] // calldata that will be used to swap the STRK and deposit the BTC
+   *     }
+   *   )
+   *
+   * @param call the call to do after the "server side" call.
+   */
+  composite(
+    inputs: { token: StarknetAddress; source: Note | Amount }[],
+    outputs: {
+      recipient: PrivateRecipient | StarknetAddress;
```
sean-starkware: how can a recipient be a StarknetAddress without a channel?

sdk/src/interfaces.ts
```diff
@@ -0,0 +1,307 @@
...
+   *     call: {
+   *       contract_address: swapHelper,
+   *       entry_point: "swap_and_deposit",
+   *       calldata: [...] // calldata that will be used to swap the STRK and deposit the BTC
+   *     }
+   *   )
+   *
+   * @param call the call to do after the "server side" call.
+   */
+  composite(
```
sean-starkware: I think we can redesign this. It can get really complicated to create the composite, and you need to walk on "eggshells" to avoid breaking anything every time you need to do something new that is not a simple transfer to a single person. As of this state, I think that a wallet will always use the composite, even for a simple transfer.

How about a builder pattern that represents a pipe of basic actions? That way, it is more convenient for both the user and the implementor, as it opens the door to good validations, optimizations, etc. Basic actions can be "transfer(like what you have above)", "call", "splitNote", etc.

Examples based on the examples you gave:

transfer 10 STRK to Alice and 20 Eth to Bob using 3 notes:
```ts
pipe.transfer(Alice, STRK, [note1], 10n)
        .transfer(Bob, Eth, [note2, note3], 20n)
        .execute()
```

swap 10 STRK for BTC using a 20 STRK note
```ts
pipe.transfer(SwapHelper, STRK, [note1], 10n, selfChannel) //no need for manual remainder
        .call(swapHelper, 'swap_and_deposit', [...])
        .execute()
```
Maybe execute is not the right word since its not really execute, maybe `prove` or something else

You can come up with something more accurate and flexible, but i think you get the point.

We can still have a `composite` action in the pipe, as a "custom action" that can be used in a special case of something that cant be done as a standard "action", but we can expect it to be rarely used (if at all)

We can look at everything as pipes of actions, taking advantage of multicalls, no?

sdk/src/internal.ts
```diff
@@ -0,0 +1,199 @@
...
+  Witness,
+  WitnessSerde,
+  channelBrand,
+  witnessBrand,
+} from "./interfaces.js";
+
+// Base nonce class with shared logic and methods.
+class Nonce {
+  created?: BlockNumber;
+  readonly nonce0: number;
```
sean-starkware: Maybe I should not go that deep in the review, but we can give better name then nounce0 and nounce1, and represents what they actually mean

sdk/src/internal.ts
```diff
@@ -0,0 +1,199 @@
...
+  increment(): this {
+    const Ctor = this.constructor as new (nonce0: number, nonce1: number) => this;
+    return new Ctor(this.nonce0, this.nonce1 + 1);
+  }
+
+  /**
+   * Finds the nonce with the lowest `created` value (skipping those without one)
+   * and returns its increment. Returns undefined if no nonces have `created`.
+   */
+  static next<T extends Nonce>(this: { new (...args: unknown[]): T }, nonces: T[]): T | undefined {
```
sean-starkware: the `this` part is not needed, since you create new instance via calling `increment`, and not via `this`. Yeah, it will not affect runtime, but I still don't favor doing things that we don't need

sdk/src/internal.ts
```diff
@@ -0,0 +1,199 @@
...
+    const oldest = withCreated.reduce((min, n) => (n.created! < min.created! ? n : min));
+    
+    // TODO: add a check the oldest is old enough (at least 10 blocks old)
+
+    return oldest.increment() as T;
+  }
+}
+
+// Factory that creates a subclass of Nonce with a specific static MAX.
+function withMax<TMax extends number>(max: TMax) {
```
sean-starkware: i guess the need for those subclasses, is the fact that you only use `max` at ctor time. how about storing the max as a class variable in the `Nonce` class, then you will give it to the new instance, created by the `increment` function. That way, no need for subclasses.

You will lose the `ChannelNonceClass` and the `TokenNonceClass` and the "explicitness" of the fact that those class has a preconfigured "max", but its a fair tradeoff for a much simpler type system.

sean-starkware: If you do keep this type system, TMax should not extend number, no need to over-engineer this

sean-starkware (commented):
I ignored the tests, since it looks like it uses some non existing classes(?)

However, i fail to understand how you are planning to only keep the "external" Witness and Channel without doing some evil castings inside/outside the implementations.

sdk/src/interfaces.ts
ittaysw: It was just an attempt to show a starknet.js like interface. Will be used by the implementation.

sdk/src/interfaces.ts
ittaysw: If there's no remainder, it is not required.

sdk/src/interfaces.ts
ittaysw: same reply

sdk/src/interfaces.ts
ittaysw: withdrawal

sdk/src/interfaces.ts
ittaysw: I like the suggestion. I need to see the final contracts design (they just finished refactoring) to see if the entire 'composite' notion makes sense.

sdk/src/internal.ts
ittaysw: I'll think of better names.

sdk/src/internal.ts
ittaysw: right

sdk/src/internal.ts
ittaysw: if it's a class variable, it is shared by all classes. But we have 2 values (10 and 100). If it's an instance variable, it seems a bit of shame to include it in every instance. Probably not a big deal, but that is what i tried to avoid.

sean-starkware: If you look at the 10 and 100 as instances and not child classes, you would say, "I currently have two instances of 10 and 100". Which is fine, but your class implementation supports any number anyway.

If you want it to support only 10 and 100, then don't accept a number in the ctor, accept an ENUM of tokens or channel. (which I don't recommend doing)

My point is, if you make the class to support any `max`, then be ok with supporting any `max`, and be comfortable with storing it as a class variable. especially when you have the benefit of transferring it to the new instance created by "next".

sdk/src/interfaces.ts
sean-starkware: My point is, that the user doesn't always know if there will be a reminder or not. Sending the selfChannel when you didn't need to, is better than having an error because you needed to send selfChannel but you did not know.

sdk/src/interfaces.ts
ittaysw: Note that the builder pattern doesn't completely remove the need for composite since it's not clear how it can be used to take a single note and transfer it to two people (without needing to meddle with "in flight" remainders and the possibility of transfer without input notes). I can do a builder of spend() and create() that allows something like spend(note1).spend(note2).create(alice, notea).create(alice, noteb), not sure it is more intuitive than a composite method with two arrays.

sean-starkware: unless you implement a transfer to multiple destinations by `[recipient, amount][]`
which makes sense anyway

sdk/src/interfaces.ts
ittaysw: I don't believe in this "doesn't know". We're talking about wallets and money. I want to avoid the need to discover the self channel (it changes as transactions are executed on SN)

sdk/src/internal.ts
ittaysw: i'm saying you can't store "any max" as a *class* variable without the trick of creating a class constructor. Let's discuss f2f if you disagree.

sdk/src/interfaces.ts
ittaysw: then i have implemented half of the composite function anyway, so simplified little.

sdk/src/internal.ts
sean-starkware: sure let meet f2f, im just saying that its ok that the 10 and 100 will not reflect in the interface

sdk/src/interfaces.ts
```diff
@@ -35,14 +43,19 @@ export type Witness = {
 export type WitnessSerde<B extends BlobT = string> = Serde<Witness, B>;
 
 export type Note = {
-  readonly id: string;
+  readonly id: Note.Id;
```
sean-starkware: what am i missing? why not use `BigNumberish` directly? and it you want to mask it under an ID type, lets just create a type of `NoteId = BigNumberish`. I dont think we need namespace here

sdk/src/interfaces.ts
```diff
@@ -198,27 +198,27 @@ export interface PrivateTransfers {
...
-    token: StarknetAddress,
-    inputs: Note[],
-    amount?: Amount,
-    selfChannel?: Channel
-  ): Promise<PrivateInvocationResult>;
+  transfer(params: {
+    token: StarknetAddress;
+    inputs: Note[];
+    recipient: PrivateRecipient | Self;
+    amount?: Amount | Open;
```
sean-starkware, line 180: Doesn't it make the amount non-optional?

ittaysw:
> However, i fail to understand how you are planning to only keep the "external" Witness and Channel without doing some evil castings inside/outside the implementations.

There's code that serializes and deserializes.

sdk/src/interfaces.ts
ittaysw: it's a marker. if the amount is not given then the total of the inputs is transferred.

sdk/src/interfaces.ts
ittaysw: Maybe too much typescripting here. Wanted it to be like a property of Node

ittaysw:
> I ignored the tests, since it looks like it uses some non existing classes(?)

Don't ignore tests. Fixed

sdk/src/interfaces.ts
sean-starkware: In the doc you said that the marker says "use all remaining input", it means something else than "total of the inputs"?

sdk/src/internal.ts
```diff
@@ -0,0 +1,205 @@
...
+    readonly channelKey: ChannelKey,
+    readonly nonce: TokenNonce
+  ) {
+    super();
+  }
+}
+
+export const witnessSerde: WitnessSerde = {
+  encode(ctx) {
+    const internal = ctx as InternalWitness;
```
sean-starkware: This is the evil casting im talking about. In most languages, this kind of dynamic casting would raise a red flag..

Thing is, if the witness is created by our sdk anyway, why we want this internal/external thing? what is the harm of having a single witness? if the properties are private and readonly anyway

ittaysw:
* **#62** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/62?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/62?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

## How to use the Graphite Merge Queue

Add the label <kbd>merge-queue</kbd> to this PR to add it to the merge queue.

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/starkware-libs?ref=merge-queue-instructions-in-stack-comment&prId=62).

<h2></h2>
<sub>

An organization admin has enabled the [Graphite Merge Queue](https://graphite.dev/docs/graphite-merge-queue) in this repository.

</sub>

<sub>

Please do not merge from GitHub as this will restart CI on PRs being processed by the merge queue.

</sub>

<h2></h2>

sdk/src/interfaces.ts
ittaysw: i fixed the comment. leftover from a previous marker attempt.

sdk/src/internal.ts
ittaysw: Yeah, was an overkill. Simplified.

sdk/src/interfaces.ts
sean-starkware: not critical, but I'm still preferring a simple `BigNumberish` or `type NoteId = BigNumberish`, to avoid usage of old native TypeScript features like namespaces.

Feel free to resolve if you do not agree

sdk/src/internal.ts
```diff
@@ -109,26 +106,27 @@ export const channelSerde: ChannelSerde = {
...
+  /** @internal */ readonly nonce: TokenNonce;
+
+  constructor(channelKey: ChannelKey, nonce: TokenNonce) {
+    this.channelKey = channelKey;
+    this.nonce = nonce;
   }
 }
 
 export const witnessSerde: WitnessSerde = {
-  encode(ctx) {
```
sean-starkware: Instead of doing it that way, would you consider having a kind of a `serializable` interface that witness and channel will implement, that does the `encode` and `decode`? That way, instead of weakly marking the properties as Internal, you can make them private. 

wdyt?

ittaysw: I would still keep them internal (don't end up in index.d.ts). The current interface allows for several serialization results (if classes implement the interface, then need to pass some marker for the results and have a 'switch', and can't add other serializations externally - not a big deal for this project, but anyway)

sdk/src/interfaces.ts
ittaysw: your wish is my command. made NoteId.

sdk/src/internal.ts
ittaysw: I rewrote using a simple class based approach. Also made it more robust.

sean-starkware (approved):

PR #394, opened 2026-02-03
fix: skip open channel when already exists

dhruvkelawala:
## Summary
- avoid emitting OpenChannel actions when the channel key is already known
- prevent NON_ZERO_VALUE reverts on transfers that include auto-setup

## Testing
- npm run check (sdk)

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/394)


ittaysw:
Can you provide the code that failed before the fix and is successful after? It seems like channel opening (setup action on the builder) was requested when a channel was already open. This is a user mistake. If you rely on the registry / discovery, you can avoid calling setup and these will be added by the compiler.

PR #423, opened 2026-02-05
fix: use next outgoing channel index

dhruvkelawala:
## Summary
- add regression test for deposit->transfer sequence that reuses outgoing channel index
- tighten mock pool to enforce sequential outgoing channel indices
- fix compiler to compute next outgoing channel index via discovery

## Problem
`OpenChannel.index` was using `seenRecipients.size - 1` (batch order), not the actual next outgoing channel slot. If a sender already has an outgoing channel (e.g., self-channel from deposit), the next transfer reuses index 0 and reverts with `NON_ZERO_VALUE`.

## Solution
When opening channels, discover all outgoing channels first to get the current count, then use that as the starting index and increment per new channel.

## Testing
- `pnpm -C sdk test` (all pass)

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/423)

### DIFF
--- a/sdk/src/internal/compiler.ts
+++ b/sdk/src/internal/compiler.ts
@@ -80,6 +80,13 @@ export class ActionCompiler {
     const registry = options?.registryConst ? this.cloneRegistry(registry_) : registry_;
     const recipientsNeeded = this.getRecipientsNeeded(actions);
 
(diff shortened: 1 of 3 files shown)


ittaysw:
The problem was index tracking. see https://github.com/starkware-libs/starknet-privacy/pull/442

PR #493, opened 2026-02-16
feat(sdk): add native ClientAction::Swap support

dhruvkelawala:
## Summary

This change enables the SDK to produce native `ClientAction::Swap` actions that the Cairo contract already supports, rather than decomposing swaps into `Withdraw + CreateOpenNote + FollowupCall`.

The `FollowupCall` was being filtered before reaching the contract (in `serialization.ts`), so swaps silently failed against real contracts (only worked with the mock pool).

## Changes

- **`interfaces.ts`**: Add `SwapAction` type and `.swap()` method to builder interface
- **`builders.ts`**: Implement `.swap()` on the builder
- **`compiler.ts`**: Add swap compilation logic in `ActionCompiler` (resolves channels, produces the native `Swap` client action)
- **`simple-private-transfers.ts`**: Update `swap()` to use native `.swap()` instead of decomposed path
- **`mock-pool-contract.ts`**: Add `Swap` case handler for tests
- **`simple-private-transfers.test.ts`**: Update test to use new `SwapAction` API

## Follow-up Fix

- **`compiler.ts`**: Align swap deposit note index with open-note creation (`index = noteNonce > 0 ? noteNonce - 1 : noteNonce`) so native swap targets the created open note instead of the next nonce slot.
- **`pool-simulator.ts`**: Stop incrementing note nonce inside `handleSwap()` to match contract behavior (swap fills an existing open note rather than creating a new one).

## Why this matters

The Cairo contract has `ClientAction::Swap` as a first-class variant (phase 7). The SDK previously had no way to produce it — only the broken decomposed path. Now swaps will work against the real contract, not just the mock.

## Testing

- Build passes
- `simple-private-transfers` swap path uses native `SwapAction`
- 26 pre-existing Cairo compatibility test failures remain in `hashes.test.ts` and `encryption.test.ts` (unrelated to this change)

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/493)

### DIFF
--- a/sdk/src/interfaces.ts
+++ b/sdk/src/interfaces.ts
@@ -165,6 +165,23 @@ export type FollowupCallAction = {
   call: Call;
 };
 
+export type SwapAction = {
+  /** The SwapExecutor contract address */
+  swapExecutor: StarknetAddressBigint;
+  /** The AMM/DEX contract to call for the swap */
+  swapContract: StarknetAddressBigint;
+  /** The selector of the swap function on the AMM/DEX */
+  swapSelector: bigint;
+  /** The calldata to pass to the swap function */
+  swapCalldata: bigint[];
+  /** The token to swap from (input token) */
+  inToken: StarknetAddressBigint;
+  /** The token to swap to (output token) */
+  outToken: StarknetAddressBigint;
+  /** The amount of input tokens to swap */
+  inAmount: Amount;
+};
+
 /** Actions - context comes from registry */
 export type Actions = {
   setViewingKey?: SetViewingKeyAction;
@@ -176,6 +193,7 @@ export type Actions = {
   withdraws?: WithdrawAction[];
   surpluses?: SurplusAction[];
   followupCall?: FollowupCallAction;
+  swap?: SwapAction;
 };
 
 // ============ Auto-Discovery & Registry Types ============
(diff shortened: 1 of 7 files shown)


dhruvkelawala:
Context from devnet debugging (what failed before the follow-up fixes):

1) **Incomplete swap path with old flow (`Withdraw + CreateOpenNote + FollowupCall`)**
- Symptom: input token left the pool / reached swap executor, but output token never came back as a private note.
- Trace showed no internal call to `SwapExecutor.swap` and no `MockAMM.swap` call.
- Root cause: `FollowupCall` is client-side and filtered before Cairo actions are executed, so the on-chain path was effectively incomplete.

2) **`NOTE_NOT_FOUND` after moving to native `ClientAction::Swap`**
- Error chain ended in `deposit_to_open_note` with:
  - `0x4e4f54455f4e4f545f464f554e44 ('NOTE_NOT_FOUND')`
- Selector path:
  - `SwapExecutor.swap` (`0x15543c...`) -> `Privacy.deposit_to_open_note` (`0x3aa8af...`) -> `NOTE_NOT_FOUND`
- Root cause: swap used the next nonce slot instead of the just-created open-note index, so deposit targeted a non-existent note id.

The follow-up in this PR addresses both classes of issues: native swap path + note index alignment / simulator consistency.

PR #631, opened 2026-03-09
refactor(sdk): consistent cursors in private registry and discovery

m-kus:
## Refactor PrivateRegistry to use discovery cursors instead of storing channels directly

This change restructures the `PrivateRegistry` to hold discovery cursors (`notesCursor` and `channelCursor`) instead of directly storing channels. The cursors now contain the actual channel data and are used for incremental synchronization with reorg detection.

**Key changes:**
- Replace `registry.channels` and `registry.cursor` with `notesCursor` and `channelCursor` fields
- Move channel storage into `channelCursor.channels` 
- Add `blockId` field to `ChannelCursor` for reorg detection
- Update discovery methods to return cursor information in their responses
- Reorganize indexer discovery code into separate directory structure
- Fix incremental note discovery to properly use channel keys from previous cursor state
- Update documentation to reflect new registry structure and cursor-based discovery flow

The registry now serves as a container for discovery state rather than directly holding user data, enabling better incremental sync and reorg recovery capabilities.

This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/631)

PR #644, opened 2026-03-12
feat(sdk): streamline registry management and discovery strategies

m-kus:
### 1. Registry owns its mutation logic

`PrivateRegistry` is now a class with three methods — `applyDiscoveredNotes`, `applyDiscoveredChannels`, `applyExecuteResult` — instead of a plain type alias that every caller mutated ad-hoc. Merge semantics (missing vs refresh, cursor merging) live in one place.

### 2. Compilation produces data, not side effects

`ExecuteResult` no longer returns the same registry reference pre-mutated during compilation. Instead it returns a `RegistryUpdate` data object that the caller applies explicitly — typically after on-chain tx confirmation. This eliminates `registryConst` (a workaround for eager mutation) and gives the caller control over when optimistic state is committed.

### 3. Discovery and note selection are separate phases

The compiler's `resolveNotes` was a single async function that discovered notes, updated the registry, selected notes, and created surplus actions. Discovery is now a dedicated phase before note resolution. `resolveNotes` is synchronous — it only selects from what's already in the registry.

### 4. Token/recipient filters are separated from discovery strategies

The old API conflated filtering with discovery levels: `"all"` meant "discover all tokens regardless of balance", `"total-only"` meant "don't return channels, just count them", and `RecipientsFilter` was a union of `T[] | "all" | "total-only"`. Now discovery level (`"missing"` | `"refresh"`) controls only the cursor strategy, and token/recipient arrays control filtering orthogonally. The compiler extracts action tokens as a filter; callers pass explicit arrays or omit for all.

## What was wrong

**Scattered registry mutations.** The compiler, AbstractPrivateTransfers, and PoolSimulator each had their own copy of registry update logic. Adding a field to the registry required updating every caller. Cursor merge logic was duplicated with no single source of truth for consistency.

**Eager mutation during compilation.** `PoolSimulator.updateRegistry()` unconditionally wrote new channels and notes into the registry during `compile()`, before the tx was submitted. If the tx failed, the registry was corrupted. `registryConst: true` existed as a workaround (clone before mutation), adding API surface for what should have been a non-issue. The returned `registry` on `ExecuteResult` was the same reference passed in — confusing and impossible to diff.

**Entangled discovery and selection.** `resolveNotes` was a 100-line async function mixing balance calculation, token filtering, discovery calls, registry mutation, note selection, and surplus creation. The `"all"` discovery level added another branch to decide "discover all tokens regardless of balance" vs "only tokens with deficits".

**Double outgoing sync call.** When not all recipients had open channels, the compiler made a second `discoverChannels("total-only")` call just to fetch the total count — an extra round-trip now unnecessary because the cursor already contains the total.

**Incremental sync was broken.** Outgoing channel discovery never seeded the cursor from the registry, so every call did a full rescan. Notes had a similar problem — discovery results were written to the registry but the cursor wasn't properly merged, so incremental updates re-fetched already-known data. Neither `discoverNotes` nor `discoverChannels` applied their results back to the registry — callers had to do it manually, and the compiler didn't.

**Cursor not forwarded.** `discoverNotes` consumed the cursor internally and didn't return it, so callers couldn't access `totalNoteCounts` for UI display.

### `PrivateRegistry` class (`channel.ts`)

```typescript
export class PrivateRegistry {
  notesCursor?: NotesCursor;
  channelCursor?: ChannelCursor;
  notes: AddressMap<Note[]>;

  applyDiscoveredNotes(level, notes, cursor): void;  // missing=append, refresh=replace + cursor merge
  applyDiscoveredChannels(cursor): void;              // merge channel cursor in-place
  applyExecuteResult(update: RegistryUpdate): void;   // optimistic post-tx state (no cursor changes)
}
```

### `RegistryUpdate` data type

```typescript
export type RegistryUpdate = { channels: AddressMap<Channel>; notes: AddressMap<Note[]> };
```

Returned by `PoolSimulator.createRegistryUpdate()` and surfaced on `ExecuteResult.registryUpdate`. Plain data — inspectable, deferrable, applied only when the caller decides to.

### Compiler

- Discovery split into dedicated `discoverNotes` phase; `resolveNotes` is now sync (select + surplus only)
- Single `discoverChannels` call (total from cursor, no `"total-only"` second call)
- Returns `registryUpdate: RegistryUpdate` instead of mutated registry
- Removed `cloneRegistry`, `allOpen`, `registryConst`
- Token filter extracted from actions via `getActionTokens()`

### `AbstractPrivateTransfers`

- `discoverNotes`/`discoverChannels` accept `{ registry, level, tokens/recipients }` and update registry via its methods
- `discoverNotes` returns `cursor: NotesCursor`
- Removed `RecipientsFilter` union — recipients is `StarknetAddress[] | undefined`

### Other files

| File | Change |
|------|--------|
| `interfaces.ts` | Re-export `PrivateRegistry` class and `RegistryUpdate` from `channel.ts`. Remove old type alias, `createEmptyRegistry`, `registryConst`, `RecipientsFilter`. |
| `pool-simulator.ts` | `updateRegistry(registry)` -> `createRegistryUpdate(): RegistryUpdate` (data, no mutation) |
| `private-transfers.ts` | Destructure `registryUpdate` instead of `registry` |
| `simple-private-transfers.ts` | Object literal -> `new PrivateRegistry()` |
| `mocknet.ts` | `executeOutside(result): PrivateRegistry` -> `executeOutside(result, registry?): void` |
| `README.md` | State management section rewritten: transaction discovery, post-tx update, periodic refresh |

### Registry creation

```diff
- import { createEmptyRegistry } from "starknet-sdk";
- const registry = createEmptyRegistry();
+ import { PrivateRegistry } from "starknet-sdk";
+ const registry = new PrivateRegistry();
```

Object literals no longer satisfy the type:
```diff
- const registry: PrivateRegistry = { channels: new AddressMap(), notes: new AddressMap() };
+ const registry = new PrivateRegistry();
```

### Post-execution update

```diff
- const { registry } = await transfers.execute(actions, { registry: myRegistry });
- // registry === myRegistry, already mutated
+ const { registryUpdate } = await transfers.execute(actions, { registry: myRegistry });
+ // myRegistry is NOT mutated — apply only after tx confirmation:
+ myRegistry.applyExecuteResult(registryUpdate);
```

Same for `createProofInvocation`:
```diff
- const { invocation, registry } = await transfers.createProofInvocation(actions, options);
+ const { invocation, registryUpdate } = await transfers.createProofInvocation(actions, options);
```

### Remove `registryConst`

No longer needed — execution doesn't mutate the registry.
```diff
  await transfers.execute(actions, {
    registry,
-   registryConst: true,
  });
```

### Mocknet

```diff
- const updatedRegistry = mocknet.executeOutside(result);
+ mocknet.executeOutside(result, registry);
```

### Discovery (direct calls)

`discoverNotes` — new params shape, returns cursor:
```diff
- const { notes } = await transfers.discoverNotes({ cursor: registry.cursor, tokens: [STRK] });
+ const { notes, cursor } = await transfers.discoverNotes({ registry, level: "missing", tokens: [STRK] });
```

`discoverChannels` — positional `recipients` arg replaced with params object:
```diff
- const { channels, total } = await transfers.discoverChannels("all", { cursor });
+ const { channels, total } = await transfers.discoverChannels({ registry, level: "missing" });

- const { channels } = await transfers.discoverChannels([bob, alice], { cursor });
+ const { channels } = await transfers.discoverChannels({ registry, level: "missing", recipients: [bob, alice] });
```

### `AutoDiscoveryOptions.notes`

`"all"` is removed — token filter is now derived from the actions automatically:
```diff
  autoDiscover: {
-   notes: "all",
+   notes: "refresh",
    channels: "missing",
  }
```

### `DiscoveryProviderInterface` (custom discovery providers)

`discoverChannels` signature changed:
```diff
  discoverChannels(
    address: StarknetAddressBigint,
    viewingKey: ViewingKey,
-   recipients: RecipientsFilter,
+   recipients: StarknetAddressBigint[] | undefined,
    params?: { cursor?: ChannelCursor }
- ): Promise<{ timestamp: BlockIdentifier; channels?: AddressMap<Channel>; total?: number }>;
+ ): Promise<{ timestamp: BlockIdentifier; channels?: AddressMap<Channel>; total?: number; cursor: ChannelCursor }>;
```

Return type now includes `cursor: ChannelCursor`.

### Removed exports

| Export | Replacement |
|--------|-------------|
| `createEmptyRegistry()` | `new PrivateRegistry()` |
| `RecipientsFilter` | `StarknetAddress[]` or `undefined` |
| `ExecuteResult.registry` | `ExecuteResult.registryUpdate: RegistryUpdate` |
| `ExecuteOptions.registryConst` | removed (no-op) |
| `AutoDiscoveryOptions.notes: "all"` | `"refresh"` (tokens derived from actions) |

This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/644)

### DIFF
--- a/sdk/src/internal/compiler.ts
+++ b/sdk/src/internal/compiler.ts
@@ -19,15 +19,14 @@
 import type {
   Actions,
   Amount,
+  DiscoveryLevel,
   DiscoveryProviderInterface,
   ExecuteOptions,
-  Note,
-  PrivateRegistry,
   StarknetAddressBigint,
   ViewingKey,
   Warning,
 } from "../interfaces.js";
-import { Channel, createEmptyRegistry, WarningCode } from "../interfaces.js";
+import { Channel, PrivateRegistry, WarningCode, type RegistryUpdate } from "../interfaces.js";
 import { AddressMap, AdvancedMap, toBigInt } from "../utils/index.js";
 import type { ClientAction } from "./client-actions.js";
 import { PoolSimulator } from "./pool-simulator.js";
@@ -43,7 +42,8 @@ import { ReorgError } from "./indexer/index.js";
 
 export type CompileResult = {
   clientActions: ClientAction[];
-  registry: PrivateRegistry;
+  /** Optimistic state update (new notes and channels). Apply via `registry.applyExecuteResult()` after tx confirmation. */
+  registryUpdate: RegistryUpdate;
   warnings: Warning[];
 };
 
@@ -110,8 +110,7 @@ export class ActionCompiler {
   }
 
   private async compileOnce(actions: Actions, options?: ExecuteOptions): Promise<CompileResult> {
-    const registry_ = options?.registry ?? createEmptyRegistry();
-    const registry = options?.registryConst ? this.cloneRegistry(registry_) : registry_;
+    const registry = options?.registry ?? new PrivateRegistry();
     const recipientsNeeded = this.getRecipientsNeeded(actions);
 
     // Phase 1: Resolve recipient channels
@@ -122,8 +121,15 @@ export class ActionCompiler {
       recipientsNeeded
     );
 
-    // Phase 2: Resolve notes (discover and/or auto-select)
-    await this.resolveNotes(actions, registry, options);
+    // Phase 2a: Discover notes (update registry)
+    const notesDiscoveryLevel = options?.autoDiscover?.notes;
+    if (notesDiscoveryLevel) {
+      const actionTokens = this.getActionTokens(actions);
+      await this.discoverNotes(registry, notesDiscoveryLevel, actionTokens);
+    }
+
+    // Phase 2b: Resolve notes (select from registry, handle surpluses)
+    this.resolveNotes(actions, registry, options);
 
     debugLog("compiler", "compile", "post resolveNotes", registry?.notes?.size, actions);
 
@@ -137,7 +143,7 @@ export class ActionCompiler {
 
     return {
       clientActions,
-      registry: pool.updateRegistry(registry),
+      registryUpdate: pool.createRegistryUpdate(),
       warnings: this.checkWarnings(clientActions),
     };
   }
@@ -183,6 +189,16 @@ export class ActionCompiler {
       }
     }
     return recipientsNeeded;
+  }
+
+  private getActionTokens(actions: Actions): StarknetAddressBigint[] {
+    const tokens = new Set<bigint>();
+    for (const d of actions.deposits ?? []) tokens.add(d.token);
+    for (const u of actions.useNotes ?? []) tokens.add(u.token);
+    for (const w of actions.withdraws ?? []) tokens.add(w.token);
+    for (const c of actions.createNotes ?? []) tokens.add(c.token);
+    for (const s of actions.surpluses ?? []) tokens.add(s.token);
+    return [...tokens];
   }
 
   private createPool(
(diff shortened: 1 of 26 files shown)


m-kus:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/644?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#644** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/644?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/644?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#631** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/631?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#653](https://github.com/starkware-libs/starknet-privacy/pull/653) <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/653?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

sdk/src/internal/compiler.ts
```diff
@@ -137,7 +143,7 @@ export class ActionCompiler {
 
     return {
       clientActions,
-      registry: pool.updateRegistry(registry),
+      registryUpdate: pool.createRegistryUpdate(),
```
ittaysw, line 146: if at the end a discovery update is returned from the pool, why is there a need for PrivateRegistry to have the apply methods?

sdk/src/internal/contract-discovery.ts
```diff
@@ -416,7 +413,7 @@ export class ContractDiscoveryProvider extends AbstractDiscoveryProvider {
   async discoverChannels(
     address: StarknetAddressBigint,
     viewingKey: ViewingKey,
-    recipients: RecipientsFilter,
+    recipients: StarknetAddressBigint[] | undefined,
```
ittaysw: 'recipients?'

sdk/tests/internal/integration.test.ts
```diff
@@ -219,20 +228,22 @@ describe("Private Transfers Integration", () => {
       // Deposit 100n, transfer only 30n to Bob -> 70n surplus with NO explicit surplusTo
       mocknet.executeOutside(await bob.build().register().execute());
 
-      const result = mocknet.executeOutside(
+      const registry = new PrivateRegistry();
+      mocknet.executeOutside(
         await alice
-          .build(AUTO_ALL)
+          .build({ ...AUTO_ALL, registry })
```
ittaysw: why need to explicitly pass an empty registry?

sdk/tests/internal/indexer-discovery.test.ts
```diff
@@ -502,7 +511,7 @@ describe("IndexerDiscoveryProvider", () => {
       const provider = createProvider();
       const fetchMock = mockFetchJson({ body: outgoingSyncResponse() });
 
-      await provider.discoverChannels(USER_ADDRESS, VIEWING_KEY, "all");
+      await provider.discoverChannels(USER_ADDRESS, VIEWING_KEY, undefined);
```
ittaysw: the use of "all" was to purposefully label the meaning. undefined is confusing here I think

ittaysw (changes requested):
I don't think we should change the API surface at this point (returning an update instead of a registry), since it'll require wallets to change their implementation & tests. We didn't get negative feedback on the current API, so it seems to work for them. 

Removing the 'total' discovery, now that the discovery service supports it, can be a separate PR. IIRC I wanted to keep it separate so it may be skipped some times. Not sure if this holds here. This will also allow to review better as this PR is huge.

Separation / cleanup in the compiler can probably be a separate PR as well. 

Changing "all" to undefined can be a separate PR.

sdk/src/internal/compiler.ts
m-kus: Not sure what do you mean
The apply method is to apply the update to the registry

sdk/tests/internal/indexer-discovery.test.ts
m-kus: Moved to the optional params (similarly to discoverNotes/tokenFilter), no need for explicit undefined
Why undefined: closely match the Option<Address[]> semantics used by the discovery

sdk/src/internal/contract-discovery.ts
m-kus: Moved to optional params

sdk/tests/internal/integration.test.ts
m-kus: No need, removed

PR #713, opened 2026-04-22
feat(discovery): surface note creation block to clients

m-kus:
# feat(discovery): surface note creation block to clients

## Why

Notes in the privacy pool aren't immediately spendable: clients must wait for the **10-block maturity rule** before referencing a note in a withdrawal, or the contract rejects the call. Today the SDK has no way to know how mature a note is — `discovery-core` reads note slots via `getStorageAt` returning only the slot value, throwing away the per-slot block metadata the RPC could have included for free.

The result is that the SDK can't gate spend attempts on maturity. It either:
- over-rejects (waits longer than necessary, e.g. blocking on the chain head before trying), or
- under-rejects (sends the call and gets a contract-side rejection that the user has to interpret).

The `Note.created?: BlockNumber` field already exists on the SDK's `Note` type as a stub — it's just never populated because the wire format doesn't carry the data.

## What

Switch the notes-discovery storage reads from value-only to value-plus-block, plumb the block number through `DecryptedNote`, and populate `Note.created` in the SDK.

The Starknet RPC supports an `IncludeLastUpdateBlock` flag on `getStorageAt` that piggybacks the slot's last-update block onto the same call (no extra round-trip, no extra cost). Note slots are write-once in the privacy-pool design, so `last_update_block` of a note slot equals the block in which the note was created.

The infra to fetch this metadata already exists in the trait (`IViews::get_notes_batch_with_block`) and is already used by the history flow — the notes-discovery flow just hadn't adopted it yet.

## Changes

**`discovery-core`**
- `DecryptedNote` gains `block_number: u64`.
- `process_note_batch`, `exponential_probe`, and `bisect_boundary` switch to the `_with_block` storage variants. The probe cache now stores `(note_id, packed, last_update_block)` so cached probe-hits flow through to the linear scan without re-fetching.
- New `IViews::get_note_with_block` default method (single-note variant for the bisect path).
- Removed now-unused `IViews::get_note` and `IViews::get_notes_batch` — sole callers were the three sites being switched.
- `COST_NOTE` and `COST_NOTE_PROBING` doc-comments updated to reference the new method names. Numeric cost is unchanged: `IncludeLastUpdateBlock` rides on the same `getStorageAt` call.

**`discovery-service`**
- No code change. `IncomingSyncResponse.notes` directly serializes `DecryptedNote`, so the new field auto-propagates onto the wire as `"block_number": <u64>`.
- Spec `06-api-design.md`: updated the JSON example for the notes response and added a paragraph explaining the field's source (slot's `last_update_block` via the `IncludeLastUpdateBlock` flag) and purpose (10-block maturity gate).

**SDK**
- `ApiIncomingNoteInfo` gains `block_number: number`.
- `convertIncomingNotes` populates `Note.created` from it.
- `CHANGELOG.md` entry under `## Unreleased` → `### Added`.

## Tests

- Updated existing `test_decrypt_note_*` to pass and assert `block_number`.
- New `test_discover_notes_propagates_block_number`: inserts two open notes at distinct blocks via `MockBackend::insert_with_block`, runs `discover_notes_paginated`, asserts each `DecryptedNote.block_number` matches the inserted block.
- All existing notes / last-note-index tests continue to pass after the cache type change.

## Notes

- The change keeps the wire format additive — older SDK clients ignore the new `block_number` field and continue to function. New SDK code that reads `Note.created` requires the updated discovery-service.
- `block_number = 0` is a possible-but-degenerate value (e.g. notes against a `MockBackend` that wasn't given a block via `insert_with_block`). On real chain data it will always be the genuine creation block. Consumers of `Note.created` should treat `0` as "unknown" if they want to be defensive against pre-deployment fixture data, but this isn't a concern for the indexer flow.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/713)

### DIFF
--- a/crates/discovery-core/src/discovery/last_note_index.rs
+++ b/crates/discovery-core/src/discovery/last_note_index.rs
@@ -9,6 +9,7 @@
 
 use std::collections::HashMap;
 
+use starknet_core::types::StorageResult;
 use starknet_types_core::felt::Felt;
 
 use crate::discovery::cursor::SubchannelCursor;
@@ -38,7 +39,8 @@ pub fn boundary_budget(max_note_log_index: u32) -> usize {
 /// fully or makes no progress (budget too small).
 ///
 /// Returns `(probe_cache, has_more)`:
-/// - `probe_cache`: index → (note_id, packed_amount) from probe hits (for incoming scan).
+/// - `probe_cache`: index → (note_id, storage_result) from probe hits (for incoming scan).
+///   The `StorageResult` carries both the packed value and the slot's last-update block.
 /// - `has_more`: `true` if budget was insufficient and another call is needed.
 ///
 /// On success, sets `cursor.total_n_notes`. Callers read the result via
@@ -50,7 +52,7 @@ pub async fn find_last_note_index<S: IViews>(
     cursor: &mut SubchannelCursor,
     max_note_log_index: u32,
     budget: &IoBudget,
-) -> Result<(HashMap<u64, (Felt, Felt)>, bool), DiscoveryError> {
+) -> Result<(HashMap<u64, (Felt, StorageResult)>, bool), DiscoveryError> {
     if cursor.total_n_notes.is_some() {
         return Ok((HashMap::new(), false));
     }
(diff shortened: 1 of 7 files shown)


m-kus:
* **#713** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/713?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/713?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Yoni-Starkware (approved):
:lgtm:

> @Yoni-Starkware reviewed 7 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/713#-:-Oqt3k7F006llSXQfT5j:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on m-kus).

PR #719, opened 2026-04-26
feat(sdk): accept proof user without account

### LINKED ISSUE #718: ProofInvocationFactory bypasses account-level signing, breaks smart wallet signature validation
## Summary

`ProofInvocationFactory.create()` calls `user.signer.signTransaction()` directly, bypassing the account abstraction signing path. This produces raw `[r, s]` Stark signatures that fail `is_valid_signature` validation for smart wallet accounts (e.g. Argent smart accounts, multisig) which expect account-formatted signatures.

## Current behavior

In `src/internal/private-transfers.ts`, the account is reduced to a `ProofUser`:

```ts
const invocation = await this.params.proofInvocationFactory.create(
  { address: this.params.account.address, signer: this.params.account.signer, viewingKey },
  this.params.poolContractAddress, clientActions, details,
);
```

Then in `ProofInvocationFactory.create()` (`src/internal/proof-invocation-factory.ts`):

```ts
const signature = await user.signer.signTransaction(
  [{ contractAddress: poolAddressHex, entrypoint: "compile_actions", calldata: executeViewCalldata }],
  { walletAddress: poolAddressHex, cairoVersion: "1", ...detailsWithNonce }
);
```

This calls the raw signer directly. For basic accounts where `is_valid_signature` just does ECDSA verify, raw `[r, s]` works. For smart wallets, the account contract expects a structured signature format (e.g. signer type enum, public key, guardian co-signature, length prefix).

## Observed failure

When using an Argent smart account (with guardian/cosigner), the prover calls `is_valid_signature` on the user's account contract with the raw `[r, s]` signature and gets `invalid-signature-length`:

```
selector: 0x028420862938116cb3bbdbedee07451ccc54d4e9412dbef71142ad1980a30941  (is_valid_signature)
```

The proof invocation itself shows only 2 signature felts:
```json
"signature": ["0x3605...", "0x3611..."]
```

## Why this is not a simple fix

The SDK cannot just call `account.signTransaction(calls, details)` instead, because:

1. Most account implementations (starknet.js `Account`, Argent `SmartAccount`, etc.) rebuild signer details internally via `buildInvocationSignerDetailsPayload()`, which hardcodes `walletAddress: this.address` (the user account address).
2. For proof invocations, the transaction hash must use `sender_address = poolAddress`, not the user account address.
3. The prover validates the signature against this pool-sender hash via `is_valid_signature` on the user account.

So the current approach correctly controls what hash gets signed (`walletAddress: poolAddress`), but fails to let the account format the resulting signature.

## Design tension

- The **SDK** needs to control the signing details (specifically `walletAddress: poolAddress`) because the proof invocation sender is the privacy pool, not the user account.
- The **account** needs to own signature formatting because `is_valid_signature` is account-contract-specific.

Currently the SDK owns both, which only works for basic ECDSA accounts.

### A. Signature formatting callback

Add an optional callback that post-processes the raw signature:

```ts
createPrivateTransfers({
  account,
  formatProofSignature?: (rawSignature: Signature, account: Account) => Promise<Signature>,
  ...
})
```

The SDK signs with `signer.signTransaction()` as today, then passes the result through the formatter. Smart wallets provide the formatter to wrap signatures in their account-specific format.

### B. Proof signer override

Add an optional signer that replaces `account.signer` for proof invocations:

```ts
createPrivateTransfers({
  account,
  proofSigner?: SignerInterface,
  ...
})
```

The smart wallet provides a signer whose `signTransaction()` internally handles owner + guardian signing and returns the account-formatted signature, while respecting the `walletAddress` from the passed details.

### C. Account-level signing with detail passthrough

Refactor so the SDK calls `account.signTransaction(calls, details)` but ensures the account respects the provided `walletAddress` from details rather than overriding it. This may require coordination with starknet.js or wallet account implementations.

## Environment

- SDK version: 0.14.2
- Affected account types: Any account where `is_valid_signature` expects more than raw `[r, s]` (Argent smart accounts, multisig accounts, accounts with guardian/cosigner)
- Works correctly for: Basic Argent accounts without multi-signer class hashes

## Related code

- `sdk/src/internal/proof-invocation-factory.ts` — where raw signer is called
- `sdk/src/internal/private-transfers.ts` — where account is reduced to ProofUser
- `sdk/src/testing/mock-proving.ts` — validates signature via `is_valid_signature` on user account

dhruvkelawala:
## Summary

Allows `createPrivateTransfers()` to accept a minimal proof user directly:

```ts
createPrivateTransfers({
  user: { address, signer },
  viewingKeyProvider,
  provingProvider,
  discoveryProvider,
  poolContractAddress,
})
```

This removes the need to provide a full `Account` when the SDK only needs the user's address and a signer for proof invocations.

Closes #718

## Problem

`ProofInvocationFactory` signs proof invocations through a `SignerInterface`. Plain ECDSA accounts work, but smart wallets often need account-formatted signatures for `is_valid_signature` validation — for example signer type metadata, owner/guardian composition, or other wallet-specific wrapping.

A full `Account` is not needed for proof invocation construction. The SDK needs:

- user address
- signer used for the proof invocation
- viewing key from the viewing key provider

## Solution

Add a preferred `user: { address, signer }` input shape and make `PrivateTransfers` depend internally on that minimal user identity.

For backwards compatibility, existing callers can still pass:

- `account`
- `account + proofSigner`

When `account + proofSigner` is used, the factory converts it to the same internal user shape:

```ts
user = {
  address: account.address,
  signer: proofSigner ?? account.signer,
}
```

This addresses Ittay's direction to avoid requiring an `Account`, while preserving the original `proofSigner` hatch for existing integrations.

## Changes

- Add `PrivateTransfersUser` interface
- Add `CreatePrivateTransfersParams` union:
  - preferred `user: { address, signer }`
  - backwards-compatible `account` with optional `proofSigner`
- Update `PrivateTransfers` to store `user` instead of `account/proofSigner`
- Update mocknet helper to construct the minimal user shape
- Add test coverage for minimal-user factory input and proof invocation signing

## Tests

- `cd sdk && npm run build -- --pretty false` ✅
- `cd sdk && npm run lint` ✅
- `cd sdk && npx vitest run --coverage.enabled=false tests/factory.test.ts tests/proof-signer.test.ts` ✅

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/719)


ittaysw:
Why don't you pass an instance of your account implementation that will sign the way you expect?

dhruvkelawala:
That's how I expected it to work. But sdk destructs account into { address, signer } and calls signer.signTransaction.

If it used account.signTransaction, it would have worked as Account is responsible for merging and formatting signatures from multiple signers.

Explained in the issue in more details

ittaysw:
So the solution is to not destruct the account. Do you want to submit a PR for that (and test it works for you) or you prefer me to do it?

Yoni-Starkware (commented):
Here's a PR. @dhruvkelawala works for you?

> @Yoni-Starkware reviewed all commit messages and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/719#-:-OrNL4bt2UO_pxp6ApKW:b-tjoaca)* status: 0 of 3 files reviewed, all discussions resolved.

Yoni-Starkware (commented):
https://github.com/starkware-libs/starknet-privacy/pull/727

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/719#-:-OrNLB241217ZKFia1s4:be5th0d)* status: 0 of 3 files reviewed, all discussions resolved.

dhruvkelawala:
Posted a longer write-up on #727 — short version: neither PR resolves the architectural issue cleanly because the SDK calls `signer.signTransaction(...)` directly, bypassing `Account.signTransaction` where wallets like ArgentX put their signature wrapping (owner + guardian merge).

#727 doesn't change the bypass — it just reads `account.signer` fresh.
#719 (this PR) is an explicit hatch around it.

The principled fix would be: SDK calls `account.signTransaction(calls, details)` and Accounts are documented to honor caller-supplied `walletAddress`. That removes the need for both #719 and #727's relevance to this issue.

Discussion on #727 — let's converge there.

sdk/src/internal/private-transfers.ts
```diff
@@ -29,6 +29,7 @@ export class PrivateTransfers extends AbstractPrivateTransfers {
     private readonly params: {
       account: Account; // the user account (for signing)
```
ittaysw: might as well change this to just an address

ittaysw (commented):
after reading your comment on #727, i agree with your approach.

dhruvkelawala:
Addressed Ittay's clarification/direction here: the SDK now accepts a minimal `user: { address, signer }` directly, so consumers don't need to provide a full Account for proof invocations.

Existing `account` and `account + proofSigner` callers remain backwards-compatible; internally `PrivateTransfers` now normalizes both paths to the same minimal proof user shape.

Verification:

- `cd sdk && npm run build -- --pretty false`
- `cd sdk && npm run lint`
- `cd sdk && npx vitest run --coverage.enabled=false tests/factory.test.ts tests/proof-signer.test.ts`

sdk/src/factory.ts
```diff
@@ -57,22 +58,57 @@ function isDiscoveryProviderConfig(
...
   discoveryProvider: DiscoveryProviderInterface | DiscoveryProviderConfig;
   poolContractAddress: StarknetAddress;
-}): PrivateTransfersInterface {
+}
+
+export type CreatePrivateTransfersParams = CreatePrivateTransfersBaseParams &
+  (
+    | {
+        /** Minimal user identity. Prefer this over passing a full Account. */
+        user: PrivateTransfersUser;
```
ittaysw: can't it be anything that has an address property or a signer property? Then an Account can fit as well.

ittaysw (changes requested):

ittaysw (approved):

Yoni-Starkware (approved):
:lgtm:

> @Yoni-Starkware reviewed 9 files and all commit messages, made 1 comment, resolved 2 discussions, and dismissed @ittaysw from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/719#-:-Ot9DvbxBp2Oc6_fN3O5:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhruvkelawala).

Yoni-Starkware:
Merged this: https://github.com/starkware-libs/starknet-privacy/pull/760

Yoni-Starkware:
So closing this one.

PR #727, opened 2026-04-29
fix(sdk): use account reference in ProofUser instead of extracted signer

Yoni-Starkware:
Replace
   `address + signer` fields in `ProofUser` with a single `account: Account` reference. The factory now    
  reads `account.signer` at signing time rather than from a pinned extracted reference, so custom Account 
  subclasses (Braavos, ArgentX, hardware wallets) that override the signer property are respected.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/727)

### DIFF
--- a/sdk/src/internal/private-transfers.ts
+++ b/sdk/src/internal/private-transfers.ts
@@ -67,7 +67,7 @@ export class PrivateTransfers extends AbstractPrivateTransfers {
     // Create invocation for proving
     const details = await this.params.provingProvider.getDefaultDetails();
     const invocation = await this.params.proofInvocationFactory.create(
-      { address: this.params.account.address, signer: this.params.account.signer, viewingKey },
+      { account: this.params.account, viewingKey },
       this.params.poolContractAddress,
       clientActions,
       details
--- a/sdk/src/internal/proof-invocation-factory.ts
+++ b/sdk/src/internal/proof-invocation-factory.ts
@@ -6,12 +6,7 @@
  * - Mock implementation: passes through client actions directly
  */
 
-import type {
-  BigNumberish,
-  CallResult,
-  SignerInterface,
-  V3InvocationsSignerDetails,
-} from "starknet";
+import type { Account, BigNumberish, CallResult, V3InvocationsSignerDetails } from "starknet";
 import type { constants } from "starknet";
 import { CallData, ETransactionVersion, hash, stark } from "starknet";
 
@@ -62,8 +57,7 @@ export function getDefaultProofDetails(
  * Minimal user info needed for creating a proof invocation.
  */
 export interface ProofUser {
-  address: BigNumberish;
-  signer: SignerInterface;
+  account: Account;
   viewingKey: BigNumberish;
 }
 
@@ -126,7 +120,7 @@ export class ProofInvocationFactory implements ProofInvocationFactoryInterface {
   ): Promise<ProofInvocation> {
     const cairoActions = serializeClientActions(clientActions);
     const callDataCompiler = new CallData(PrivacyPoolABI);
-    const userAddress = toBigInt(user.address);
+    const userAddress = toBigInt(user.account.address);
     const poolAddressHex = toHex(poolAddress);
 
     const executeViewCalldata = callDataCompiler.compile("compile_actions", [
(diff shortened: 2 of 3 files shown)


dhruvkelawala:
Thanks @Yoni-Starkware. Tested it — and we're stuck on a deeper issue that neither this PR nor #719 fully resolves cleanly. Want to lay it out before we pick a path:

### The constraint

The pool verifies `is_valid_signature(hash, sig)` where `hash` is computed with `walletAddress = poolAddress` (the pool is the sender of the proof transaction, even though the user owns the signature). So the SDK MUST be able to control `walletAddress` when it asks the user's account to sign.

### What the SDK does today

Calls `signer.signTransaction(calls, details)` directly, with `details.walletAddress = poolAddress`. This works for plain ECDSA accounts.

But it bypasses any account-level signature work — for ArgentX smart accounts, owner + guardian signatures merge in our `Account` subclass, not in the signer. The signer alone returns raw `[r, s]`. Pool rejects with `argent/invalid-signature-length`. Multisig and hardware wallets that orchestrate at the Account level would hit the same shape of issue.

### What #727 does for this

Reads `account.signer` fresh instead of pinning it. **Doesn't change the bypass** — still calls `signer.signTransaction(...)`. So smart-wallet signature wrapping is still skipped.

#727 is a fine generalization for cases where wallets override `signer` itself (custom signer wrappers), but our merge logic lives at Account level, not signer level. We don't want to refactor it down — Account-vs-signer separation is intentional and clean.

### What #719 does for this

Adds a `proofSigner?` hatch so the wallet can inject a custom signer just for proofs. We can build a `PrivacyProofSigner` that does the owner+guardian merge inside the signer, and the SDK uses that instead of `account.signer` when supplied. Works, but it's a workaround — every wallet with Account-level wrapping has to provide one.

### The principled fix

The right shape is: SDK calls something on `account` that lets the Account apply its own signature scheme, while the SDK still controls `walletAddress`.

**Catch:** `starknet.js` `Account` (v10) does not have a public `signTransaction(calls, details)` method. It has `execute()` (too high-level — submits the tx), `signMessage()` (SNIP-12 typed data, different scheme), and `buildInvocation()` (calls `this.signer.signTransaction()` internally — no Account-level wrapping). Our `BaseStarknetAccount.signTransaction()` is a method we added in our subclass; the SDK can't rely on it being there for arbitrary wallets.

So the principled fix needs one of:

1. **starknet.js adds a public `Account.signTransaction(calls, details)`** that honors caller-supplied `walletAddress` and calls `this.signer.signTransaction()` internally. Subclasses (ArgentX, multisig, hardware) override to add their wrapping. The SDK calls this. — Cleanest, requires upstream change.

2. **The SDK documents a contract**: the `account` argument must implement `signTransaction(calls, details)` that honors `walletAddress`. Wallets with Account-level wrapping subclass; plain ECDSA users provide a thin wrapper. — No starknet.js change, but introduces a non-standard contract.

3. **Keep the SDK calling `signer.signTransaction()`** and provide an explicit hook for wallets where wrapping lives at Account level. — That's `proofSigner?` from #719. Workaround, but small and unblocks us today.

### Where we are

- We don't want to restructure our SmartAccount to move merge logic into a signer wrapper. That's a real refactor across web + mobile.
- We don't love `proofSigner?` either — it's a workaround.
- The principled fix needs a starknet.js change (option 1) or a normative SDK contract (option 2).

What's your appetite for option 1 / 2? Otherwise we keep `proofSigner` as the explicit hatch (option 3) and revisit later.

cc @ittaysw

ittaysw:
I agree with the approach in #719: allow to provide a signer and avoid the need to provide an Account.

ittaysw (changes requested):
I think we should close this PR

dhruvkelawala:
Thanks @ittaysw — happy to lock in option 3 (`proofSigner` hatch).

Quick clarification on "avoid the need to provide an Account": is your preference (a) keep #719 essentially as-is — `account` stays required, `proofSigner?` overrides only the signer used for proof invocation — or (b) something more — e.g. SDK accepts `{ address, signer }` directly and `account` becomes optional / dropped?

Default to (a) unless you say otherwise — it's a smaller diff and fully backwards-compatible. Will rebase #719 against current main and tighten the JSDoc/test coverage. Plan to land #727 alongside as a standalone refactor.

PR #759, opened 2026-05-18
fix(sdk): support OHTTP on React Native

dhruvkelawala:
## Summary
- switch SDK OHTTP HPKE suite to noble-backed X25519/HKDF/AES-GCM primitives
- add a React Native-compatible SubtleCrypto fallback for ohttp-ts internal HMAC and AES-GCM operations
- add noble OHTTP interop coverage for request/response roundtrips and copied response buffers

## Context
React Native's crypto.subtle implementation does not support the HMAC/AES-GCM importKey paths used internally by ohttp-ts while decapsulating OHTTP responses. This caused mobile OHTTP response decryption to fail even though request encapsulation worked.

This keeps the existing synthetic inner request origin behavior from main, so gateway path prefixes are not leaked into encrypted inner requests.

## Validation
- npm run build
- npm run lint
- npm run test:fast -- --coverage.enabled=false
- live OHTTP route sanity check returns expected API 400 for dummy viewing key instead of transport/decryption failure

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/759)

### DIFF
--- a/sdk/package-lock.json
+++ b/sdk/package-lock.json
@@ -9,6 +9,9 @@
       "version": "0.14.2-rc.5",
       "license": "ISC",
       "dependencies": {
+        "@noble/ciphers": "^2.2.0",
+        "@noble/hashes": "^2.2.0",
+        "@panva/hpke-noble": "^1.1.0",
         "@starknet-io/starknet-types-0101": "npm:@starknet-io/types-js@~0.10.2",
         "ohttp-ts": "^0.3.0",
         "starknet": "^10.0.0-beta.6",
@@ -787,6 +790,18 @@
         "@jridgewell/sourcemap-codec": "^1.4.14"
       }
     },
+    "node_modules/@noble/ciphers": {
+      "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/@noble/ciphers/-/ciphers-2.2.0.tgz",
+      "integrity": "sha512-Z6pjIZ/8IJcCGzb2S/0Px5J81yij85xASuk1teLNeg75bfT07MV3a/O2Mtn1I2se43k3lkVEcFaR10N4cgQcZA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">= 20.19.0"
+      },
+      "funding": {
+        "url": "https://paulmillr.com/funding/"
+      }
+    },
     "node_modules/@noble/curves": {
       "version": "1.7.0",
       "resolved": "https://registry.npmjs.org/@noble/curves/-/curves-1.7.0.tgz",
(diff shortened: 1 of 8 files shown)


dhruvkelawala:
Addressed Codex review feedback in `7fdcd5a`:

- Moved the install guard so early construction before `crypto.subtle` exists does not permanently disable the fallback.
- HMAC fallback now preserves the requested hash and supports SHA-256/SHA-384/SHA-512; unsupported hashes rethrow the native import error instead of silently using SHA-256.

Validation rerun:
- `npm run build`
- `npm run lint`
- `npm run test:fast -- --coverage.enabled=false`

sdk/src/internal/ohttp-client.ts
```diff
@@ -52,6 +54,7 @@ export class OhttpClient {
     private readonly gatewayUrl: string,
     options?: { relayUrl?: string; publicKeyConfig?: Uint8Array }
   ) {
+    installOhttpWebCryptoFallback();
```
ittaysw: maybe have clients that need it call it rather than being called always? they can then call it once instead of per client ctor and maybe earlier.

sdk/package-lock.json
```diff
@@ -9,6 +9,9 @@
       "version": "0.14.2-rc.5",
       "license": "ISC",
       "dependencies": {
+        "@noble/ciphers": "^2.2.0",
+        "@noble/hashes": "^2.2.0",
```
ittaysw, line 13: noble/hashes already appears as transitive dependency (1.6), any way to eliminate?

sdk/src/internal/ohttp-webcrypto-fallback.ts
```diff
@@ -0,0 +1,181 @@
...
+ *
+ * TODO: Remove this once ohttp-ts accepts injectable HKDF/AEAD primitives.
+ */
+let didInstall = false;
+
+export function installOhttpWebCryptoFallback(): void {
+  if (didInstall) return;
+
+  const subtle = globalThis.crypto?.subtle;
+  if (typeof subtle?.importKey !== "function") return;
```
ittaysw, line 22: i think this means that tests don't cover this fallback.

ittaysw (changes requested):

dhruvkelawala:
Addressed Ittay's review in `ae9ac40`:

- The fallback is no longer installed from `OhttpClient` constructor.
- The SDK now exports `installOhttpWebCryptoFallback()` for clients that need it to call once during app/bootstrap setup, before creating OHTTP clients.
- Added direct tests for the fallback:
  - install can be retried when `crypto.subtle` becomes available later
  - HMAC fallback respects SHA-384
  - AES-GCM encrypt/decrypt fallback works

On the `@noble/hashes` dependency: `@panva/hpke-noble` does bring noble packages transitively, but this SDK imports `@noble/hashes/*` and `@noble/ciphers/*` directly in the fallback, so keeping them as direct dependencies avoids relying on transitive package layout. Once upstream `ohttp-ts` pluggable crypto lands, this fallback and these direct imports should be removed.

Validation rerun:
- `npm run build`
- `npm run lint`
- `npm run test:fast -- --coverage.enabled=false` — 214 tests passed

dhruvkelawala:
Closing this SDK-owned fallback approach. We validated OHTTP on React Native by moving the missing WebCrypto subset polyfill to the mobile client instead, where `react-native-quick-crypto` is available. The SDK should not own a React Native-specific `crypto.subtle` monkey-patch.

We'll keep tracking the cleaner long-term fix via the upstream ohttp-ts pluggable response crypto PR: https://github.com/thibmeu/ohttp-ts/pull/7

PR #858, opened 2026-06-29
fix(sdk): paginate total-only outgoing discovery (fixes 257th-recipient NON_ZERO_VALUE)

Yoni-Starkware:
## Summary

Fixes a real off-by-one that makes **a private transfer to the 257th distinct recipient fail with `NON_ZERO_VALUE`** (while the first 256 work) — the "shield rush" finding against private-payroll, which uses this SDK.

When the SDK opens a channel to a **new** recipient it needs the sender's current outgoing-channel count, which it reads via `discoverChannels("total-only")`. That path did a **single** `/v1/sync/outgoing_state` request. But the discovery service caps a single request at `max_channels` (default **256**, `discovery-core/.../cursor.rs:34`) and only sets `total_n_channels` once it reaches the **sentinel** (`outgoing_channels.rs:140-143`). So for a sender with ≥256 existing channels the request breaks at the cap with `has_more=true` and returns **no total**.

That `undefined` total then propagated:

- `compiler.ts:128` → `createPool(..., total)` → `new PoolSimulator(..., totalChannels ?? 0)` (`compiler.ts:191`) — the `?? 0` turns it into `nextChannelIndex = 0`
- `compiler.ts:296` → the OpenChannel action gets `index: 0`
- on-chain, opening a channel at index 0 hits the **write-once** guard on the already-occupied slot 0 → `errors::NON_ZERO_VALUE` (`privacy.cairo:902`)

**Boundary, confirmed from code:** a sender with `N=255` channels (indices 0–254) discovers all 255 + sentinel in one request → `total=255` → 256th channel opens at index 255 ✅. With `N=256` (indices 0–255) the request caps before the sentinel → `total=undefined` → index 0 → collision on the 257th ❌. Existing recipients keep working because re-paying them skips the count query (`allOpen` short-circuit, `compiler.ts:531-533`).

## Fix

Make `total-only` **paginate to the sentinel** in a `do/while` loop, pruning complete channels each round (so the cursor frees slots and discovery advances past the cap) — exactly mirroring the existing `discoverChannels("all")` / `discoverNotes` loops. The count is then authoritative regardless of channel count. If the service ever reports `channel_discovery_complete` without a `total_n_channels`, we now **throw** instead of silently coercing to 0.

A brand-new sender still correctly returns `total = 0` (sentinel at index 0 → service sets `Some(0)`, verified by the Rust test `test_paginated_zero_channels_sets_total`), so the throw guard never false-positives.

## Tests

`sdk/tests/internal/indexer-discovery.test.ts` (29 pass):
- **≥256 pagination**: incomplete first page (no total) → sentinel page → `total=257`, asserts 2 requests + block_ref/cursor handoff. Fails against the old single-request code (1 request, `undefined`).
- **new-sender `total=0`** without throwing.
- **contract-violation throw** when complete-but-no-total.

## Verification

- `npm run lint` (prettier + eslint + tsc): clean
- `npm run test:fast`: 240/240 pass
- CHANGELOG updated under `## Unreleased` → `### Fixed`

An independent adversarial review (subagent) confirmed: the bug is real with the exact 256/257 boundary, the fix is correct/terminating with no stall, the regression test is meaningful (verified failing on old code, passing on fix), and there are no must-fix issues.

## Note / follow-up

The root cause is SDK-side, but `total-only` now walks all of the sender's channels (and the service processes their subchannels/notes) just to get a count. A cheaper follow-up would be to return `total` from the main `discoverChannels` walk (it already reaches the sentinel) and drop the second query, and/or add a dedicated outgoing channel-count view server-side. Out of scope here to keep the fix minimal.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/858)

### DIFF
--- a/.claude/rules/code-style.md
+++ b/.claude/rules/code-style.md
@@ -114,6 +114,12 @@ Apply these guidelines when writing or reviewing code in this codebase.
 - Add comments where the reasoning isn't obvious from context
 - Focus on decisions, constraints, and non-obvious requirements
 
+### Describe the present, not the history
+- A comment explains how the code behaves now, not what it used to do or which bug it fixed
+- Don't narrate past errors, prior implementations, or the diff — that belongs in commit messages / PR descriptions
+- *Bad:* `// previously returned undefined, which the caller turned into index 0 and a collision`
+- *Good:* `// the service sets the total only once discovery reaches the sentinel, so paginate to completion`
+
 ### No visual separator comments
 - Don't use decorative comment lines to delineate sections (e.g., `// -----------`, `// =========`, `// ***`)
 - If a file needs visual structure, that's a signal to split into separate modules or use doc comments on items
--- a/sdk/CHANGELOG.md
+++ b/sdk/CHANGELOG.md
@@ -12,6 +12,17 @@
   the sentinel (e.g. on-chain `ContractDiscovery`) still fall back to the
   dedicated count query.
 
+### Fixed
+
+- `discoverChannels("total-only")` now paginates outgoing channel discovery to
+  the sentinel instead of issuing a single request. A sender with at least
+  `max_channels` (default 256) outgoing channels caps the first page before the
+  sentinel, so the service returned no `total_n_channels`; the count surfaced as
+  `undefined` and the compiler turned it into channel index `0`, producing a
+  `NON_ZERO_VALUE` write-once collision when opening a channel to the next
+  (e.g. 257th) recipient. The count is now derived once discovery completes, and
+  a missing total after completion throws instead of being silently coerced.
+
 ## 0.14.3-RC.1
 
 ### Added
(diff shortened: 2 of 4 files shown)


ittaysw (approved):

Yoni-Starkware (commented):
> @Yoni-Starkware reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/858)* status: 3 of 4 files reviewed, all discussions resolved (waiting on avi-starkware).

Yoni-Starkware (commented):
> @Yoni-Starkware reviewed 1 file.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/858)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on avi-starkware).

PR #866, opened 2026-07-01
fix(discovery-core): paginate /v1/history gap withdrawal scan to prevent 500s

Yoni-Starkware:
## Problem

`POST /v1/history` returned `500 INTERNAL_ERROR` (`InsufficientBudget`) for accounts whose most-recent note is far behind chain head. Server logs showed the real cause:

```
history: fetch_transactions done num_transactions=0 history_complete=false budget_remaining=9962
WARN  Insufficient I/O budget: needed 10510, available 9962
```

## Root cause

On a fresh history request the gap withdrawal scan (`fetch_aggregated_withdrawal_events`) resolved its upper bound to the snapshot head and charged the whole `[note+1, head]` range as **one indivisible** I/O cost — `ceil(span/1024) * 10`, consumed *before* the RPC. For a ~1.05M-block gap that single charge (~10510) exceeds the entire per-request budget (10000), so `try_consume` failed, the request 500'd, and the cursor never advanced → **permanent** failure for that account. The same flaw hit any account with a >~1M-block gap *between* notes, one page later.

The gap scan exists because a **full** withdrawal (spends a note, creates no change note) emits a `Withdrawal` event in a block where the user has no note, so only a block-range scan surfaces it.

## Fix — gap-first, budget-chunked scan

`/v1/history` is an incremental, newest-first scroll, which rules out reordering by spend-block (that would require knowing all spends up front). So we keep the block-range scan and **bound it**:

- `BufferedNoteScanner::peek_next_block` — returns the next note block **without** draining it.
- Per iteration: scan the gap `(K, upper]` **first**, top-down, in budget-bounded windows (`IoBudget::consume_up_to`), advancing `cursor.begin_block_number` per window. A far-behind account now paginates (200 + forward progress) instead of 500ing; a wide empty gap is absorbed in ~1 page and the cursor advances.
- Only once the gap is fully covered is block `K` drained and its events fetched. Scanning strictly **above** `K` before draining keeps `K` out of the gap range, so a budget failure mid-gap can't make a later page re-scan (and double-count) the note block's withdrawals.

**No API/SDK/cursor changes** (the cursor wire shape is unchanged; `begin_block_number` just advances to mid-gap values). Existing `to_addr` withdrawal attribution is preserved. Also removes the now-dead `DiscoveryError::CostOverflow` variant and updates the history spec.

### Behavior notes

- A page may return **few or zero** transactions while still advancing the cursor through a long inactive gap — clients must keep paginating until `history_complete` (the SDK's `fetchHistory` already exposes this; no change needed).
- Attribution is unchanged: withdrawals to fresh/unlinkable addresses remain unattributed (a pre-existing limitation; fixing it needs global spend knowledge, which is incompatible with the incremental scroll — out of scope).

## Testing

- `cargo fmt --check`, `cargo clippy --all-targets` (0 warnings), `cargo test` — discovery-core **157** + discovery-service **71** pass.
- New regression tests: `far_behind_account_paginates_instead_of_500` (reproduces the exact bug → now 200 across 2 pages) and `note_block_withdrawal_not_double_counted_across_pages` (gap-first invariant). Existing budget/pagination tests pass unchanged (per-page budget accounting is identical).
- `snforge test` — 320 pass (no contract change).

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866)


avi-starkware (commented):
> @avi-starkware+AGNT made 16 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866)* status: 0 of 5 files reviewed, 15 unresolved discussions (waiting on avi-starkware and Yoni-Starkware).

___
*[`crates/discovery-core/src/history/transactions.rs` line 45 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgF9F9maDflqxFI7W:-P0XgF9F9maDflqxFI7X:b5ayzy0) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L45)):*
> ```XML
> ///    via [`fetch_aggregated_block_events`].
> ///
> /// Budget is consumed incrementally; the scan stops at the first iteration that
> ```

Non-blocking. This still says every budget-exhausted iteration leaves the cursor at the frontier so the next page continues. The second commit added a case that returns `Err(InsufficientBudget)` instead, which maps to a 500 (`api/types.rs:119`) with no next page to continue on. Worth one clause.
___
*[`crates/discovery-core/src/history/transactions.rs` line 71 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgJrtB7wtc1yud4M_:-P0XgJrtB7wtc1yud4Ma:b-i14enn) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L71)):*
> ```XML
> 
>     loop {
>         if transactions.len() >= max_transactions {
> ```

Question, pre-existing so not this PR's problem, but it's the same loop the second commit set out to remove.

`max_transactions: 0` passes validation and breaks here before any work. `max_transactions` is a required, un-defaulted `u32` (`api/types.rs:322`), and `validate_bound` only rejects values above the max, so `0` gets through. Three consecutive probe pages:

```
attempt=1 page=Ok(0) complete=false cursor=Some(100)
attempt=2 page=Ok(0) complete=false cursor=Some(100)
attempt=3 page=Ok(0) complete=false cursor=Some(100)
```

A client paginating until `history_complete`, which `specs/06-api-design.md:353` (added here) tells it to do, loops forever. Reject `max_transactions == 0` in `validate_history_cursor` with a 400?
___
*[`crates/discovery-core/src/history/transactions.rs` line 136 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgKxc7CXfEWT661Pr:-P0XgKxc7CXfEWT661Ps:b4quikw) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L136)):*
> ```XML
>     );
> 
>     if result.is_empty() {
> ```

Question, and it interacts with the drain issue at `:251`.

`start_bound` is captured per iteration rather than per page (`:169`, inside `process_next_block`), so the `Err` branch at `:243` can fire on iteration 2 or later of a single page. This gate then swallows it whenever the page already collected something.

So this line decides 200 vs 500, not the guard, and nothing in the diff or the suite pins it. Is that the intent?
___
*[`crates/discovery-core/src/history/transactions.rs` line 169 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgC_PCE51oLcfadEK:-P0XgC_PCE51oLcfadEL:b3th4pr) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L169)):*
> ```XML
>     // The cursor bound at entry; compared later to detect whether this
>     // iteration advanced the cursor (only the gap scan moves it).
>     let start_bound = cursor.begin_block_number;
> ```

Non-blocking. `start_bound` looks redundant. At `:242` the check `cursor.begin_block_number == start_bound` is exactly `upper == note_block`: if the gap was non-empty we either returned `Halted`, or set the cursor to `Some(gap_floor - 1) == Some(note_block)`, which can't equal `start_bound`.

Dropping the variable and its comment, then writing `if upper == note_block {`, kept 158/158 with no new warnings.
___
*[`crates/discovery-core/src/history/transactions.rs` line 175 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgD9C3ijFP9OwCIh6:-P0XgD9C3ijFP9OwCIh7:bup0v71) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L175)):*
> ```XML
>     // PreConfirmed) include head-of-chain withdrawals; `block_number()` gives
>     // the matching concrete number for window arithmetic.
>     let (upper, upper_block_id) = match cursor.begin_block_number {
> ```

Non-blocking. `upper` is an adjective doing a noun's job through the rest of `process_next_block`, and it replaced the more descriptive `previous_block_number`. `.claude/rules/code-style.md:13` bans standalone adjectives as names. `upper_bound`, or `upper_block_number`?
___
*[`crates/discovery-core/src/history/transactions.rs` line 219 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XhOeg9F7WQgCqLc_L:-P0XhOeg9F7WQgCqLc_M:bgd3rks) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L219)):*
> ```XML
>         )
>         .await?;
>         cursor.begin_block_number = Some(window_bottom.saturating_sub(1));
> ```

Nothing pins this boundary. Both `Some(window_bottom)` and `Some(window_bottom.saturating_sub(2))` leave 158/158 green, and each is a real bug: a withdrawal sitting exactly at the split block comes back on two pages, or on none.

A two-page test catches both. Note at block 100, cursor upper 3000, page-1 budget `COST_NOTE + COST_EVENTS_CHUNK` puts `window_bottom` at 1977; place withdrawals at 1977 and 1976, then assert each appears exactly once across page 1 and page 2. Fails the two mutants with `left: 2  right: 1` and `left: 0  right: 1`.

Fits next to `far_behind_account_paginates_instead_of_500`.
___
*[`crates/discovery-core/src/history/transactions.rs` line 220 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0Xg8-V2CDFZENeOYuj:-P0Xg8-V2CDFZENeOYuk:bnv7yud) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L220)):*
> ```XML
>         .await?;
>         cursor.begin_block_number = Some(window_bottom.saturating_sub(1));
>         if window_bottom > gap_floor {
> ```

Non-blocking. I think this early return is dead. A partial grant leaves `remaining <= 9`, because `consume_up_to` takes `current / cost_per_item` exactly (`io_budget.rs:102`), and reaching the note step needs at least one subchannel, so `note_step_budget >= 12`. The deferral below always fires first. Its `Err` branch can't be taken either, since the cursor is now `Some(window_bottom - 1)` which is below `upper` and so never equals `start_bound`.

Replacing the condition with `if false` keeps 158/158 green, and also keeps the two boundary tests I wrote for the findings above passing. Prefer deleting it over testing it?
___
*[`crates/discovery-core/src/history/transactions.rs` line 240 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0Xg7-94zUdZGSOFMpT:-P0Xg7-94zUdZGSOFMpU:b-nhnyo4) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L240)):*
> ```XML
>     // deferring makes no progress and a same-budget retry would repeat forever
>     // on an empty response — surface the shortfall instead of looping.
>     let note_step_budget = COST_BLOCK_EVENTS_QUERY + cursor.subchannels.len() * COST_NOTE;
> ```

Non-blocking. `cursor.subchannels.len()` counts exhausted subchannels too. `fill_buffers` drops `next_index: None` entries (`notes.rs:153`) and charges only for the actives (`:160`), so the reservation is inflated by the dead ones.

On an empty-gap iteration the cursor can't move, so this returns `InsufficientBudget`, and every retry at the same budget fails identically. Probe with 1 live and 20 exhausted subchannels: real cost 12, but `needed: 52, available: 18`, repeating.

Not reachable at the default `server_budget = 10_000`. Reachable below ~522, which the clamp permits (`min_server_budget(30) = 89`, and `config.rs`'s own tests use 200).

`cursor.subchannels.iter().filter(|s| s.next_index.is_some()).count()`. Separately, should `min_server_budget` account for the history note step? It currently doesn't.
___
*[`crates/discovery-core/src/history/transactions.rs` line 248 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgG-5E6ULAq9ZToaY:-P0XgG-5E6ULAq9ZToaZ:b-9pnxiz) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L248)):*
> ```XML
>             });
>         }
>         return Ok(ScanStep::Halted);
> ```

Non-blocking. `note_step_shortfall_errors_rather_than_looping` pins the `Err` branch above (changing its condition to `if false` fails that test), but not this `Halted` branch. Turning this into an error keeps 158/158 green.

One more case in the same test covers it: whole empty gap covered, note step unaffordable, expect a 200 with the cursor moved.
___
*[`crates/discovery-core/src/history/transactions.rs` line 251 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XhM6xBYYBXuR_qjPP:-P0XhM6xBYYBXuR_qjPQ:b7wb8zl) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L251)):*
> ```XML
>     }
> 
>     let Some((block_number, block_notes)) =
> ```

The drain runs before the block-events fetch is funded. If `fetch_aggregated_block_events` fails on budget, `next_index` has already moved past the notes, and the gap withdrawal makes the page non-empty, so this returns `Ok`. The client gets a 200 plus a cursor that skipped those notes, and no later page can reach them.

Repro at this sha, 3 notes of one subchannel in block 100 plus a withdrawal at 150, budget 24:

```
page1=[150]  page2=[]  cursor=None
block-100 notes silently lost across pages
```

All 158 existing tests pass alongside it, so nothing covers this.

`note_step_budget` above doesn't prevent it. It funds one refill per subchannel, but draining several notes from the same subchannel in one block needs a refill round per note.

Suggest `budget.try_consume(COST_BLOCK_EVENTS_QUERY)?` immediately before the drain, and drop the charge inside `fetch_aggregated_block_events`. `IoBudget::reclaim` (`io_budget.rs:58`) already covers its early-return paths.

One correction to save you the argument: I first read this as a regression from the deleted `retain`/`clear` block, and that's wrong. The merge-base loses notes the same way on a two-iteration page, because that block only filtered the returned `result` vec and never restored `next_index`. So the class is pre-existing, and what this PR changes is that a single-iteration page now reaches it. That also means "restore the pre-change guarantee" isn't an option, since there wasn't one.
___
*[`crates/discovery-core/src/history/transactions.rs` line 414 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgLt87yNxamI51O2t:-P0XgLt87yNxamI51O2u:b-kkc1na) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L414)):*
> ```XML
> /// Caller guarantees `from_block <= to_block`. Returns `Err(InsufficientBudget)`
> /// only when not even one chunk is affordable (no forward progress possible).
> async fn fetch_gap_withdrawals_chunked<E: IEvents>(
> ```

Question for whoever owns the incident, not a finding.

Total per-request work is genuinely bounded now, and `far_behind_account_paginates_instead_of_500` proves the budget-exhaustion path is fixed. But the ceiling on a single `get_withdrawal_events` call is unchanged from the merge-base: same cost formula, and no per-call chunk cap independent of the budget, so `consume_up_to` can still grant the whole budget to one window.

If the production 500s came from the RPC backend timing out on a very wide range rather than from budget exhaustion, that mode isn't addressed here. Do you have the original error bodies, or the provider's `get_events` block-range limit? I couldn't settle it from the repo and didn't want to guess.
___
*[`crates/discovery-core/src/history/transactions.rs` line 1071 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgGpO24qBb770J9lm:-P0XgGpO24qBb770J9ln:b-ifam5x) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L1071)):*
> ```XML
>         .await
>         .unwrap_err();
>         assert!(
> ```

Non-blocking. The base test this replaces (`withdrawal_events_first_page_cost_uses_to_block_number`) asserted `assert_eq!(needed, COST_EVENTS_CHUNK * 2)`; this one only uses `matches!`, so nothing pins `needed` now that it's one chunk. Minor, since it's log-only at `api/types.rs:119`, but `assert_eq!(needed, COST_EVENTS_CHUNK)` is free.
___
*[`crates/discovery-core/src/history/transactions.rs` line 1298 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgHMe1_sj6zv703J6:-P0XgHMe1_sj6zv703J7:bgz28cl) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L1298)):*
> ```XML
> 
>         // Page 1: budget for the block-20 iteration only.
>         let one_block_budget = 2 * COST_NOTE + COST_BLOCK_EVENTS_QUERY + COST_EVENTS_CHUNK;
> ```

Nit, non-blocking. Fourth verbatim copy of this expression (three at the PR base). `const ONE_BLOCK_BUDGET` in `mod tests`?
___
*[`crates/discovery-core/src/history/notes.rs` line 63 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgBfb64Na_befZpzV:-P0XgBfb64Na_befZpzW:b-z048jg) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/notes.rs#L63)):*
> ```XML
>         let mut block_notes: Vec<HistoryNote> = Vec::new();
>         let mut block_number: Option<u64> = None;
>         loop {
> ```

Non-blocking. Now that `peek_next_block` exists, this loop hand-rolls what it already names. Tried it:

```rust
let max_buffered_block = self.peek_next_block(views, cursor, budget).await?;
let Some(target_block) = block_number.or(max_buffered_block) else { break };
block_number = Some(target_block);
```

4 insertions, 12 deletions, 158/158 pass, clippy clean. The peek has to stay unconditional inside the loop, since the per-iteration refill is what surfaces a second note in the same block.
___
*[`crates/discovery-core/src/history/notes.rs` line 131 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XhPgwCUsVISBO1f8w:-P0XhPgwCUsVISBO1f8x:b-bfwpi2) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/notes.rs#L131)):*
> ```XML
>         self.fill_buffers(views, &cursor.subchannels, budget)
>             .await?;
>         Ok(self.buffered_notes.values().map(|(block, _)| *block).max())
> ```

Swap `.max()` for `.min()` here and 158/158 still pass. `peek_next_block` has no direct test, and every `fetch_transactions` fixture builds a single subchannel (`transactions.rs:714`, or `vec![]` at `:1107`), so max and min return the same value in every existing test.

Peek agreeing with drain is what keeps block K out of the gap range. Two subchannel heads at different blocks plus a disagreement means the gap window includes the block about to be drained, and its withdrawals get attached twice.

`notes.rs:485` (`next_block_completes_block_across_subchannels`) already builds a multi-subchannel fixture to hang this on: assert peek equals the higher block, and that a following `next_block` drains exactly that block.
___
*[`crates/discovery-service/specs/06-api-design.md` line 299 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0Xg9F6BAFbO2MZ9Tg9:-P0Xg9F6BAFbO2MZ9TgA:bdcem13) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-service/specs/06-api-design.md#L299)):*
> ```Markdown
> - `cursor`: History cursor for pagination.
>   - `subchannels`: List of subchannels to scan. Each contains the `channel_key`, `token`, `channel_kind` (`incoming`, `outgoing`, `self_channel`), `counterparty` address, and `next_index` (next note index to read descending, `null` if exhausted).
>   - `begin_block_number`: Inclusive upper bound for the next scan window. Set to `0` on first request (server resolves from chain head). On subsequent requests, use the value from the previous response cursor — note this may land partway through a long gap above a note, since the gap is scanned in budget-bounded windows (see Cursor lifecycle).
> ```

Non-blocking, but you're already rewriting this bullet so it's cheap to fix here.

It still says set `begin_block_number` to `0` on the first request. The field is `Option<u64>` and the SDK sends `undefined` (`sdk/src/internal/history.ts:20` and `:185`), so a client following the spec literally gets `Some(0)`, which makes `upper = 0`, skips every note as above the bound, and returns an empty page with `history_complete: true`. The cursor is burned too, so re-requesting doesn't recover.

Probe at this sha, note at 100 plus withdrawal at 150: `Some(0)` gives `Ok([]) complete=true`, `None` gives `Ok([150, 100])`.

Suggest saying omit the field, or send `null`. `:345` repeats the `0` wording.

avi-starkware:
*[`crates/discovery-core/src/history/notes.rs` line 76 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0_wX9i8-vQ7pqPXVuC:-P0_wX9i8-vQ7pqPXVuD:bj87n9i) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/notes.rs#L76)):*
> ```XML
>                     None => break,
>                 },
>             };
> ```

To make it clear, this is claude's suggestion above.

Saves some lines of code, but might be less clear to read



_Suggestion:_
```XML
            let max_buffered_block = self.peek_next_block(views, cursor, budget).await?;
            let Some(target_block) = block_number.or(max_buffered_block) else { break };
            block_number = Some(target_block);
```

avi-starkware:
*[`crates/discovery-core/src/history/transactions.rs` line 71 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgJrtB7wtc1yud4M_:-P0a-bT9-jvFu8Whv5jR:b-sxae45) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L71)):*
<details><summary><i>Previously, avi-starkware+AGNT (Agent AGNT for Avi Cohen) wrote…</i></summary><blockquote>

Question, pre-existing so not this PR's problem, but it's the same loop the second commit set out to remove.

`max_transactions: 0` passes validation and breaks here before any work. `max_transactions` is a required, un-defaulted `u32` (`api/types.rs:322`), and `validate_bound` only rejects values above the max, so `0` gets through. Three consecutive probe pages:

```
attempt=1 page=Ok(0) complete=false cursor=Some(100)
attempt=2 page=Ok(0) complete=false cursor=Some(100)
attempt=3 page=Ok(0) complete=false cursor=Some(100)
```

A client paginating until `history_complete`, which `specs/06-api-design.md:353` (added here) tells it to do, loops forever. Reject `max_transactions == 0` in `validate_history_cursor` with a 400?
</blockquote></details>

I think we should return an 4xx error response when `max_transactions: 0`
This is another infinite loop case where `cursor.history_complete` can never become `true`.
See the condition that sets it to true: (it will never be met with `max_transactions` at 0)
```rust
if scanner_complete && result.len() < max_transactions {
    // ... fetch_registration ...
    cursor.history_complete = budget_error.is_none();
} else {
    cursor.history_complete = false;
}
```

avi-starkware (changes requested):
> @avi-starkware reviewed 5 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866)* status: all files reviewed, 16 unresolved discussions (waiting on avi-starkware+AGNT and Yoni-Starkware).

___
*[`crates/discovery-core/src/history/notes.rs` line 131 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XhPgwCUsVISBO1f8w:-P0a4UBh2yOLeai0RGuO:bpkkg2u) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/notes.rs#L131)):*
<details><summary><i>Previously, avi-starkware+AGNT (Agent AGNT for Avi Cohen) wrote…</i></summary><blockquote>

Swap `.max()` for `.min()` here and 158/158 still pass. `peek_next_block` has no direct test, and every `fetch_transactions` fixture builds a single subchannel (`transactions.rs:714`, or `vec![]` at `:1107`), so max and min return the same value in every existing test.

Peek agreeing with drain is what keeps block K out of the gap range. Two subchannel heads at different blocks plus a disagreement means the gap window includes the block about to be drained, and its withdrawals get attached twice.

`notes.rs:485` (`next_block_completes_block_across_subchannels`) already builds a multi-subchannel fixture to hang this on: assert peek equals the higher block, and that a following `next_block` drains exactly that block.
</blockquote></details>

TL;DR
Add unit tests for `peek_next_block`
___
*[`crates/discovery-core/src/history/transactions.rs` line 136 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgKxc7CXfEWT661Pr:-P0a6awK-8Yj-P7vT726:bxyc9yk) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L136)):*
<details><summary><i>Previously, avi-starkware+AGNT (Agent AGNT for Avi Cohen) wrote…</i></summary><blockquote>

Question, and it interacts with the drain issue at `:251`.

`start_bound` is captured per iteration rather than per page (`:169`, inside `process_next_block`), so the `Err` branch at `:243` can fire on iteration 2 or later of a single page. This gate then swallows it whenever the page already collected something.

So this line decides 200 vs 500, not the guard, and nothing in the diff or the suite pins it. Is that the intent?
</blockquote></details>

Okay this is a bit vague, but it points at something real:
A single request loops multiple iterations of `process_next_block`, but the infinite loop guard:
```rust
        if cursor.begin_block_number == start_bound {
```
checks progress each `process_next_block` iteration, and not for the entire request, so we might have: iteration 1: clearing the gap up to a note block
iteration 2: making no progress and not having enough remaining budget to drain the note_block.
iteration 2 will satisfy the condition above (meant to guard against an infinite loop), but a fresh request might have enough budget to drain the note block and might make progress.

This is a real bug and should be addressed
___
*[`crates/discovery-core/src/history/transactions.rs` line 219 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XhOeg9F7WQgCqLc_L:-P0aCa6t40ThCaHnWZ8-:b-l9941r) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L219)):*
<details><summary><i>Previously, avi-starkware+AGNT (Agent AGNT for Avi Cohen) wrote…</i></summary><blockquote>

Nothing pins this boundary. Both `Some(window_bottom)` and `Some(window_bottom.saturating_sub(2))` leave 158/158 green, and each is a real bug: a withdrawal sitting exactly at the split block comes back on two pages, or on none.

A two-page test catches both. Note at block 100, cursor upper 3000, page-1 budget `COST_NOTE + COST_EVENTS_CHUNK` puts `window_bottom` at 1977; place withdrawals at 1977 and 1976, then assert each appears exactly once across page 1 and page 2. Fails the two mutants with `left: 2  right: 1` and `left: 0  right: 1`.

Fits next to `far_behind_account_paginates_instead_of_500`.
</blockquote></details>

TL;DR
Add a test which fails if this is changed to `window_bottom` or `window_bottom.saturating_sub(2)`
___
*[`crates/discovery-core/src/history/transactions.rs` line 240 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0Xg7-94zUdZGSOFMpT:-P0aCuu450YgHLblY-YI:b-lbii4f) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L240)):*
<details><summary><i>Previously, avi-starkware+AGNT (Agent AGNT for Avi Cohen) wrote…</i></summary><blockquote>

Non-blocking. `cursor.subchannels.len()` counts exhausted subchannels too. `fill_buffers` drops `next_index: None` entries (`notes.rs:153`) and charges only for the actives (`:160`), so the reservation is inflated by the dead ones.

On an empty-gap iteration the cursor can't move, so this returns `InsufficientBudget`, and every retry at the same budget fails identically. Probe with 1 live and 20 exhausted subchannels: real cost 12, but `needed: 52, available: 18`, repeating.

Not reachable at the default `server_budget = 10_000`. Reachable below ~522, which the clamp permits (`min_server_budget(30) = 89`, and `config.rs`'s own tests use 200).

`cursor.subchannels.iter().filter(|s| s.next_index.is_some()).count()`. Separately, should `min_server_budget` account for the history note step? It currently doesn't.
</blockquote></details>

TL;DR
This calculation is not accurate because `cursor.subchannels.len()` counts subchannels that have already been fully drained as well. The correct calculation should only take active channels into account.

avi-starkware:
*[`crates/discovery-core/src/history/transactions.rs` line 220 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0Xg8-V2CDFZENeOYuj:-P0aGzVe5lWds0-x_jfj:b2160e2) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L220)):*
<details><summary><i>Previously, avi-starkware+AGNT (Agent AGNT for Avi Cohen) wrote…</i></summary><blockquote>

Non-blocking. I think this early return is dead. A partial grant leaves `remaining <= 9`, because `consume_up_to` takes `current / cost_per_item` exactly (`io_budget.rs:102`), and reaching the note step needs at least one subchannel, so `note_step_budget >= 12`. The deferral below always fires first. Its `Err` branch can't be taken either, since the cursor is now `Some(window_bottom - 1)` which is below `upper` and so never equals `start_bound`.

Replacing the condition with `if false` keeps 158/158 green, and also keeps the two boundary tests I wrote for the findings above passing. Prefer deleting it over testing it?
</blockquote></details>

TL;DR
The calculation shows that this condition can never be met.
Nevertheless, we should probably keep it as a sanity check.

avi-starkware (changes requested):
> @avi-starkware made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866)* status: all files reviewed, 16 unresolved discussions (waiting on avi-starkware+AGNT and Yoni-Starkware).

___
*[`crates/discovery-core/src/history/transactions.rs` line 248 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgG-5E6ULAq9ZToaY:-P0aHROIATIQ9WMbCzcn:b-bwxpt6) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L248)):*
<details><summary><i>Previously, avi-starkware+AGNT (Agent AGNT for Avi Cohen) wrote…</i></summary><blockquote>

Non-blocking. `note_step_shortfall_errors_rather_than_looping` pins the `Err` branch above (changing its condition to `if false` fails that test), but not this `Halted` branch. Turning this into an error keeps 158/158 green.

One more case in the same test covers it: whole empty gap covered, note step unaffordable, expect a 200 with the cursor moved.
</blockquote></details>

TL;DR
Add a test covering the `return Ok(ScanStep::Halted);` above.
___
*[`crates/discovery-core/src/history/transactions.rs` line 251 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XhM6xBYYBXuR_qjPP:-P0aIVDZ1KviLfNoPzLP:btkxtvw) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L251)):*
<details><summary><i>Previously, avi-starkware+AGNT (Agent AGNT for Avi Cohen) wrote…</i></summary><blockquote>

The drain runs before the block-events fetch is funded. If `fetch_aggregated_block_events` fails on budget, `next_index` has already moved past the notes, and the gap withdrawal makes the page non-empty, so this returns `Ok`. The client gets a 200 plus a cursor that skipped those notes, and no later page can reach them.

Repro at this sha, 3 notes of one subchannel in block 100 plus a withdrawal at 150, budget 24:

```
page1=[150]  page2=[]  cursor=None
block-100 notes silently lost across pages
```

All 158 existing tests pass alongside it, so nothing covers this.

`note_step_budget` above doesn't prevent it. It funds one refill per subchannel, but draining several notes from the same subchannel in one block needs a refill round per note.

Suggest `budget.try_consume(COST_BLOCK_EVENTS_QUERY)?` immediately before the drain, and drop the charge inside `fetch_aggregated_block_events`. `IoBudget::reclaim` (`io_budget.rs:58`) already covers its early-return paths.

One correction to save you the argument: I first read this as a regression from the deleted `retain`/`clear` block, and that's wrong. The merge-base loses notes the same way on a two-iteration page, because that block only filtered the returned `result` vec and never restored `next_index`. So the class is pre-existing, and what this PR changes is that a single-iteration page now reaches it. That also means "restore the pre-change guarantee" isn't an option, since there wasn't one.
</blockquote></details>

`process_next_block` ends in this order:

```rust
let note_step_budget = COST_BLOCK_EVENTS_QUERY + cursor.subchannels.len() * COST_NOTE;
if budget.remaining() < note_step_budget { /* defer or error */ }

let Some((block_number, block_notes)) =
    note_scanner.next_block(backend, cursor, budget).await?   // :251 drains, decrements next_index

fetch_aggregated_block_events(...).await?;                    // charges 10 FIRST, can fail here
cursor.begin_block_number = Some(block_number.saturating_sub(1));
```
The method `next_block()` decrements the `next_index` of the subchannel. `fetch_aggregated_block_events()` then does `try_consume(COST_BLOCK_EVENTS_QUERY)` _before_ reading anything, so it can fail after the `next_index` already changed. The next request will then skip the drained note.
___
*[`crates/discovery-core/src/history/transactions.rs` line 1071 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgGpO24qBb770J9lm:-P0aJnTS4GcZvKiZegxM:by7ix0x) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L1071)):*
<details><summary><i>Previously, avi-starkware+AGNT (Agent AGNT for Avi Cohen) wrote…</i></summary><blockquote>

Non-blocking. The base test this replaces (`withdrawal_events_first_page_cost_uses_to_block_number`) asserted `assert_eq!(needed, COST_EVENTS_CHUNK * 2)`; this one only uses `matches!`, so nothing pins `needed` now that it's one chunk. Minor, since it's log-only at `api/types.rs:119`, but `assert_eq!(needed, COST_EVENTS_CHUNK)` is free.
</blockquote></details>

TL;DR
Change `assert_eq!(needed, COST_EVENTS_CHUNK * 2)` to `assert_eq!(needed, COST_EVENTS_CHUNK)`

Yoni-Starkware (commented):
> @Yoni-Starkware reviewed 2 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866)* status: all files reviewed, 17 unresolved discussions (waiting on avi-starkware and avi-starkware+AGNT).

___
*[`crates/discovery-core/src/history/transactions.rs` line 71 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgJrtB7wtc1yud4M_:-P0aOV0fCeZm1f1uLiLD:biknndt) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L71)):*
<details><summary><i>Previously, avi-starkware (Avi Cohen) wrote…</i></summary><blockquote>

I think we should return an 4xx error response when `max_transactions: 0`
This is another infinite loop case where `cursor.history_complete` can never become `true`.
See the condition that sets it to true: (it will never be met with `max_transactions` at 0)
```rust
if scanner_complete && result.len() < max_transactions {
    // ... fetch_registration ...
    cursor.history_complete = budget_error.is_none();
} else {
    cursor.history_complete = false;
}
```
</blockquote></details>

Enforce max_transactions > 0
___
*[`crates/discovery-core/src/history/transactions.rs` line 220 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0Xg8-V2CDFZENeOYuj:-P0aUM2732lG8JfBiW05:b-lqpzbz) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L220)):*
<details><summary><i>Previously, avi-starkware (Avi Cohen) wrote…</i></summary><blockquote>

TL;DR
The calculation shows that this condition can never be met.
Nevertheless, we should probably keep it as a sanity check.
</blockquote></details>

Keep it
___
*[`crates/discovery-core/src/history/transactions.rs` line 207 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0aS2Pd1ghyk98LMIm1:-P0aS2Pd1ghyk98LMIm2:b45h) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L207)):*
> ```XML
>     // note block's withdrawals. `saturating_add` guards the unreachable
>     // `note_block == u64::MAX` (block numbers come from chain storage).
>     let gap_floor = note_block.saturating_add(1);
> ```





_Suggestion:_
```XML
note_block_gap_floor
```

Yoni-Starkware (commented):
> @Yoni-Starkware reviewed all commit messages, made 1 comment, and resolved 13 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866)* status: 2 of 6 files reviewed, 4 unresolved discussions (waiting on avi-starkware and avi-starkware+AGNT).

___
*[`crates/discovery-core/src/history/transactions.rs` line 136 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgKxc7CXfEWT661Pr:-P1U2xnd9vRQGO7pdz3Y:b-896fix) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L136)):*
<details><summary><i>Previously, avi-starkware (Avi Cohen) wrote…</i></summary><blockquote>

Okay this is a bit vague, but it points at something real:
A single request loops multiple iterations of `process_next_block`, but the infinite loop guard:
```rust
        if cursor.begin_block_number == start_bound {
```
checks progress each `process_next_block` iteration, and not for the entire request, so we might have: iteration 1: clearing the gap up to a note block
iteration 2: making no progress and not having enough remaining budget to drain the note_block.
iteration 2 will satisfy the condition above (meant to guard against an infinite loop), but a fresh request might have enough budget to drain the note block and might make progress.

This is a real bug and should be addressed
</blockquote></details>

Done.

Yoni-Starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866)* status: 2 of 6 files reviewed, 4 unresolved discussions (waiting on avi-starkware and avi-starkware+AGNT).

___
*[`crates/discovery-core/src/history/transactions.rs` line 251 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XhM6xBYYBXuR_qjPP:-P1Ua1TSA8XDwrGdxWY3:b-896fix) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L251)):*
<details><summary><i>Previously, avi-starkware (Avi Cohen) wrote…</i></summary><blockquote>

`process_next_block` ends in this order:

```rust
let note_step_budget = COST_BLOCK_EVENTS_QUERY + cursor.subchannels.len() * COST_NOTE;
if budget.remaining() < note_step_budget { /* defer or error */ }

let Some((block_number, block_notes)) =
    note_scanner.next_block(backend, cursor, budget).await?   // :251 drains, decrements next_index

fetch_aggregated_block_events(...).await?;                    // charges 10 FIRST, can fail here
cursor.begin_block_number = Some(block_number.saturating_sub(1));
```
The method `next_block()` decrements the `next_index` of the subchannel. `fetch_aggregated_block_events()` then does `try_consume(COST_BLOCK_EVENTS_QUERY)` _before_ reading anything, so it can fail after the `next_index` already changed. The next request will then skip the drained note.
</blockquote></details>

Done.

Yoni-Starkware (commented):
> @Yoni-Starkware resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866)* status: 2 of 6 files reviewed, 3 unresolved discussions (waiting on avi-starkware).

avi-starkware (approved):
> @avi-starkware+AGNT reviewed 4 files and all commit messages, made 3 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866)* status: all files reviewed, 2 unresolved discussions (waiting on avi-starkware).

___
*[`crates/discovery-core/src/history/transactions.rs` line 136 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgKxc7CXfEWT661Pr:-P1UoHJK0UVVy5w3ENbp:b-2j4fgm) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L136)):*

Done.
</blockquote></details>

Verified at 5ad55826: `page_start_bound` is taken once in `fetch_transactions` and the 200-vs-500 decision is the single check at its end. `later_iteration_shortfall_after_progress_is_not_an_error` covers the two-iteration case; dropping the cursor clause from that check fails it.
___
*[`crates/discovery-core/src/history/transactions.rs` line 251 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XhM6xBYYBXuR_qjPP:-P1UoIDPDf0lIniDqiK1:b-eqbmfd) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L251)):*

Done.
</blockquote></details>

Verified at 5ad55826. The rollback is stronger than the pre-funding I suggested: a refill failing mid-drain of a multi-note block is undone too, which a `try_consume` before the drain would not have covered. `drained_notes_survive_block_events_budget_failure` fails with the two `inspect_err` calls removed (`next_index` ends `None` instead of `Some(2)`).
___
*[`crates/discovery-core/src/history/transactions.rs` line 414 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XgLt87yNxamI51O2t:-P1UoIuTD52jZniZt8-6:bklssjh) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L414)):*
<details><summary><i>Previously, avi-starkware+AGNT (Agent AGNT for Avi Cohen) wrote…</i></summary><blockquote>

Question for whoever owns the incident, not a finding.

Total per-request work is genuinely bounded now, and `far_behind_account_paginates_instead_of_500` proves the budget-exhaustion path is fixed. But the ceiling on a single `get_withdrawal_events` call is unchanged from the merge-base: same cost formula, and no per-call chunk cap independent of the budget, so `consume_up_to` can still grant the whole budget to one window.

If the production 500s came from the RPC backend timing out on a very wide range rather than from budget exhaustion, that mode isn't addressed here. Do you have the original error bodies, or the provider's `get_events` block-range limit? I couldn't settle it from the repo and didn't want to guess.
</blockquote></details>

Closing. The single-call ceiling is unchanged, but it is already a committed TODO (`rpc_backend.rs:350`, pre-index privacy-pool events locally) and independent of this fix. Nothing to do here.

Yoni-Starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866)* status: all files reviewed, 2 unresolved discussions (waiting on avi-starkware).

___
*[`crates/discovery-core/src/history/transactions.rs` line 251 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/866#-P0XhM6xBYYBXuR_qjPP:-P1UoZwM8lWEJUgLds8q:b-896fix) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/0323f6303f0eac0027457344919616bba8b35cde/crates/discovery-core/src/history/transactions.rs#L251)):*
<details><summary><i>Previously, avi-starkware+AGNT (Agent AGNT for Avi Cohen) wrote…</i></summary><blockquote>

Verified at 5ad55826. The rollback is stronger than the pre-funding I suggested: a refill failing mid-drain of a multi-note block is undone too, which a `try_consume` before the drain would not have covered. `drained_notes_survive_block_events_budget_failure` fails with the two `inspect_err` calls removed (`next_index` ends `None` instead of `Some(2)`).
</blockquote></details>

Done.

PR #867, opened 2026-07-01
refactor(elliptic-proxy): default proxy-screen.mjs to /screen, add path arg

avi-starkware:
## What
`proxy-screen.mjs` now posts to `<proxy-url>/screen` and signs `/screen` by
default — the exact request the proof-interceptor makes — with an optional
positional `[path]` argument to override it.

## Why
Previously the smoke-test script POSTed the bare URL and signed only
`PROXY_SIGN_PATH` (default `/`). Reproducing the interceptor's call therefore
meant appending `/screen` to the URL **and** setting `PROXY_SIGN_PATH=/screen`
in tandem — two coupled knobs that had to agree. Defaulting to `/screen` makes
the script mirror the interceptor out of the box; the `[path]` arg is both
appended to the URL and signed, so the POSTed path and the signed path stay
consistent by construction.

## Behavior
- `node scripts/proxy-screen.mjs <url> <addr>` → POST `<url>/screen`, sign `/screen`
- `node scripts/proxy-screen.mjs <url> <addr> <path>` → POST `<url><path>`, sign `<path>`

Works for both a path-preserving gateway (`/screen` routes directly) and a Cloud
Function (name segment stripped, so the server still sees `/screen`), since the
HMAC is verified over the server-reconstructed `req.path` (`src/auth.ts`).

## Test
Verified against the deployed proxy: default → `HTTP 200` with a signature; an
override path (`/nonexistent`) → `403`, confirming the argument redirects the
request.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/867)

### DIFF
--- a/elliptic-proxy/scripts/proxy-screen.mjs
+++ b/elliptic-proxy/scripts/proxy-screen.mjs
@@ -2,19 +2,19 @@
 // Call the deployed elliptic-proxy Cloud Function for a single address.
 // Signs the request as a partner using the same HMAC scheme the proxy verifies.
 //
-// Usage:
+// Usage (path defaults to "/screen", matching the proof-interceptor):
 //   PARTNER_NAME=acme PARTNER_SECRET=<base64> \
-//     node scripts/proxy-screen.mjs <proxy-url> <address>
+//     node scripts/proxy-screen.mjs <proxy-url> <address> [path]
 //   PARTNER_NAME=acme PARTNER_SECRET_HEX=<hex> \
-//     node scripts/proxy-screen.mjs <proxy-url> <address>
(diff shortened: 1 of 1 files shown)


Yoni-Starkware (approved):
:lgtm:

> @Yoni-Starkware reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/867#-:-OwRtZzO7rCyOvS18HMb:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on avi-starkware).

PR #882, opened 2026-07-02
feat(sub_account_anonymizer): add CollectPolicy for open notes

noa-starkware:
## What

Adds a `CollectPolicy` enum on `OpenNote` so callers can choose how much of a sub-account's `token` balance is collected when settling each open note:

- **`All`** — collect the sub-account's entire `token` balance (previous, and only, behavior).
- **`Diff`** — collect only the balance *gained* during this interaction, leaving any pre-existing balance untouched.
- **`Exact(amount)`** — collect an exact `u128` amount; reverts if the sub-account doesn't have it.

## Why

Previously every open note swept the sub-account's full token balance. That conflates pre-existing funds with funds earned in the current interaction and gives callers no way to deposit a precise amount. `Diff` and `Exact` make settlement intent explicit.

## How

- `snapshot_open_notes` records each `Diff` note's pre-interaction `token` balance *before* the dapp calls run; other policies snapshot an (unused) zero.
- `collect_open_notes` computes the collected amount per policy after execution.
- New errors: `NEGATIVE_DIFF` (a `Diff` note where the interaction reduced the balance) and `INSUFFICIENT_BALANCE` (an `Exact` note exceeding the balance). `ZERO_BALANCE` still rejects any zero-amount collection.

## Testing

- `snforge test -p sub_account_anonymizer` — 18 passed, added coverage for `Diff`, `Exact`, zero-amount reverts across all policies, `NEGATIVE_DIFF`, and `INSUFFICIENT_BALANCE`.
- `snforge test -p privacy test_e2e_sub_account_anonymizer_compute_invoke` — passing (updated for the new `OpenNote` shape).

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/882)

### DIFF
--- a/packages/privacy/src/tests/test_e2e.cairo
+++ b/packages/privacy/src/tests/test_e2e.cairo
@@ -21,7 +21,7 @@ use starknet::ContractAddress;
 use starknet::account::Call;
 use starkware_utils_testing::test_utils::TokenHelperTrait;
 use sub_account_anonymizer::sub_account_anonymizer::{
-    ISubAccountAnonymizerDispatcher, ISubAccountAnonymizerDispatcherTrait,
+    CollectPolicy, ISubAccountAnonymizerDispatcher, ISubAccountAnonymizerDispatcherTrait, OpenNote,
 };
 
 // Helper: Constants for e2e testing.
@@ -1941,7 +1941,10 @@ fn test_e2e_sub_account_anonymizer_compute_invoke() {
         calldata: array![token_addr.into(), amount.into(), 0].span(),
     };
     let calls: Array<Call> = array![transfer_to_caller];
-    let open_notes: Span<(felt252, ContractAddress)> = [(note_id, token_addr)].span();
+    let open_notes: Span<OpenNote> = [
+        OpenNote { note_id, token: token_addr, collect_policy: CollectPolicy::All },
+    ]
+        .span();
     let mut invoke_data: Array<felt252> = array![];
     calls.serialize(ref invoke_data);
     open_notes.serialize(ref invoke_data);
--- a/packages/sub_account_anonymizer/src/sub_account_anonymizer.cairo
+++ b/packages/sub_account_anonymizer/src/sub_account_anonymizer.cairo
@@ -14,6 +14,17 @@ use starknet::{ClassHash, ContractAddress};
 /// The result of [`privacy_compute`]: an identity commitment that identifies a single sub-account.
 pub type IdentityCommitment = felt252;
 
+/// How much of the sub-account's `token` balance to collect for an open note.
+#[derive(Serde, Copy, Drop, PartialEq, Debug)]
+pub enum CollectPolicy {
+    /// Collect the sub-account's entire `token` balance.
+    All,
+    /// Collect only the balance gained during this interaction.
+    Diff,
+    /// Collect this exact amount.
+    Exact: u128,
+}
+
 /// An open note to settle after an interaction.
 #[derive(Serde, Copy, Drop, PartialEq, Debug)]
 pub struct OpenNote {
@@ -21,6 +32,8 @@ pub struct OpenNote {
     pub note_id: felt252,
     /// The token to deposit.
     pub token: ContractAddress,
+    /// The policy selecting how much of the sub-account's `token` balance to collect for this note.
+    pub collect_policy: CollectPolicy,
 }
 
 #[starknet::interface]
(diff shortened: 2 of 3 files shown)


remollemo (commented):
> @remollemo made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/882)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on noa-starkware and Yoni-Starkware).

___
*[`packages/sub_account_anonymizer/src/sub_account_anonymizer.cairo` line 24 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/882#-Ows7FLr4rfV2N7bsevP:-Ows7FLr4rfV2N7bsevQ:b-nsge4g) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/ff06b7fa57aa612276fdd1b5cbde743d29bb9b31/packages/sub_account_anonymizer/src/sub_account_anonymizer.cairo#L24)):*
> ```Cairo
>     /// Collect only the balance gained during this interaction.
>     Diff,
>     /// Collect this exact amount. The sub-account must have it available.
> ```

i'd trim the second sentence of hte comment.
1. obvious (can be spared)
2. badly phrased. "available" is not the best verb here.

you may however, comment somewhere about the motivation of using `Exact` this is most applicable when we cannot bundle the collection of the balance with generation of it.
if we get the balance in one tx, and we collect it in another, we cannot use Diff, but we may want not to use All. hence `Exact`



_Code quote:_
```Cairo
The sub-account must have it available.
```

remollemo (changes requested):
> @remollemo made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/882)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on noa-starkware and Yoni-Starkware).

___
*[`packages/sub_account_anonymizer/src/sub_account_anonymizer.cairo` line 308 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/882#-OwsAK_d7d1UfosTjrUz:-OwsAK_d7d1UfosTjrV-:brm1kd6) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/ff06b7fa57aa612276fdd1b5cbde743d29bb9b31/packages/sub_account_anonymizer/src/sub_account_anonymizer.cairo#L308)):*
> ```Cairo
>                 transfer_calls
>                     .append(build_transfer_call(:token, recipient: anonymizer, amount: collected));
>                 token_contract.approve(spender: privacy_contract, amount: collected);
> ```

i think this would fail in case of multiple deposits of the same token
say, i have three notes of 100 USDC
i collect all 300 USDC to the anon contract using the execute()
and I approved 100USDC 3 times.
I will manage to transfer_from (anon->privacy) once, but would not have allowance for the 2nd & 3rd times.
solution - use increase_allowance instead of approve.



_Code quote:_
```Cairo
 token_contract.approve(spender: privacy_contract, amount: collected);
```

remollemo (approved):
> @remollemo reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/882)* status: all files reviewed, 2 unresolved discussions (waiting on noa-starkware and Yoni-Starkware).

___
*[`packages/sub_account_anonymizer/src/sub_account_anonymizer.cairo` line 308 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/882#-OwsAK_d7d1UfosTjrUz:-Oww1GqwAhCpRfr6spRu:buy63ip) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/ff06b7fa57aa612276fdd1b5cbde743d29bb9b31/packages/sub_account_anonymizer/src/sub_account_anonymizer.cairo#L308)):*
<details><summary><i>Previously, remollemo wrote…</i></summary><blockquote>

i think this would fail in case of multiple deposits of the same token
say, i have three notes of 100 USDC
i collect all 300 USDC to the anon contract using the execute()
and I approved 100USDC 3 times.
I will manage to transfer_from (anon->privacy) once, but would not have allowance for the 2nd & 3rd times.
solution - use increase_allowance instead of approve.
</blockquote></details>

after talking this F2F - having limitation such as "only one note per token" should not be done using design by failure approach.
since we have generic helper contract, builders/users can come up with whatever flow they envision.
if we disallow this basic thing - it need to be clearly documented and demonstrated in tests

PR #895, opened 2026-07-07
fix(sub-account-anonymizer): doc and test multi notes per token case

noa-starkware:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/895)

### DIFF
--- a/packages/privacy/src/tests/test_e2e.cairo
+++ b/packages/privacy/src/tests/test_e2e.cairo
@@ -1993,3 +1993,139 @@ fn test_e2e_sub_account_anonymizer_compute_invoke() {
     assert!(sub_account.is_non_zero());
     assert_eq!(token.balance_of(address: sub_account), 0);
 }
+
+/// E2E: two open notes on the *same* token in one sub-account interaction is unsupported and fails.
+#[test]
+#[test_case(CollectPolicy::All)]
+#[test_case(CollectPolicy::Diff)]
+#[should_panic(expected: 'ERC20: insufficient balance')]
+fn test_e2e_sub_account_anonymizer_two_notes_same_token_fails_insufficient_balance(
+    policy: CollectPolicy,
+) {
+    let mut test: Test = Default::default();
+    let token = test.new_token();
+    let token_addr = token.contract_address();
+    let payout = 100_u128;
+    let mut user = test.new_user();
+
+    let anonymizer = deploy_sub_account_anonymizer(privacy_address: test.privacy.address);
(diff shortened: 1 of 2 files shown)


remollemo (approved):
> @remollemo reviewed 2 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/895)* status: all files reviewed, 2 unresolved discussions (waiting on noa-starkware).

___
*[`packages/privacy/src/tests/test_e2e.cairo` line 1996 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/895#-Ox-gCfVDLnf5WGnCT6O:-Ox-gCfW2ecmFsplJTmh:b-vew3n) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/68b4476dcc146a4d000c70e5e05b53f062345978/packages/privacy/src/tests/test_e2e.cairo#L1996)):*
> ```Cairo
>     assert_eq!(token.balance_of(address: sub_account), 0);
> }
> ```

this test is not making much sense.
clearly if we have one interaction in amount A and two notes that try to clear it it'd fail. 
but the failure is more subtle.

let's assume we don't have an issue with approval (i.e. assume we gave privacy infinite approbal at deployment).

the scenario a user may expect to succeed is if we have two interactions, 
1. amount := A1
2. amount := A2

then - for each of them we'd expect 2 notes with respective amounts.

now - with `EXACT` policy we'd succeed even with current code (assuming approval etc.)
With `ALL` it's obvious that we'd fail.
But we could assume DIFF would work, and it would if the diff baselines would be managed with multi in mind.

TLDR - the test should be a proper two interactions, and two notes, and not a trivial failure one.
___
*[`packages/sub_account_anonymizer/src/sub_account_anonymizer.cairo` line 72 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/895#-Ox-kabbA_98clmg-jol:-Ox-kabbA_98clmg-jom:b-7h4rv4) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/68b4476dcc146a4d000c70e5e05b53f062345978/packages/sub_account_anonymizer/src/sub_account_anonymizer.cairo#L72)):*
> ```Cairo
>     ///   selected by its [`collect_policy`](CollectPolicy) is collected from the sub-account into
>     ///   this anonymizer and recorded as a deposit. Pass at most one note per token; otherwise, the
>     ///   transaction will fail later in the privacy contract because the second approval overwrites
> ```

would



_Code quote:_
```Cairo
will
```

PR #938, opened 2026-08-03
fix(discovery-service): cap outstanding JSON-RPC requests, not connections

avi-starkware:
## What

`config.max_concurrent_requests` did not limit outstanding JSON-RPC calls. Moves the limit from the `reqwest` connector to the JSON-RPC transport, and rejects a zero limit at construction.

## Why

`RpcBackend::new` installed the limiter via `reqwest::ClientBuilder::connector_layer`. From reqwest 0.13.2's own doc comment on that method (`src/async_impl/client.rs:2412`):

> Adds a new Tower `Layer` to the base connector `Service` **which is responsible for connection establishment**.

The layer wraps `Service<Unnameable, Response = Conn, ...>` — the connect step. A request served by an already-pooled connection never passes through it. And because TCP connect is near-instant relative to request latency, even a *fresh* connection clears the permit during the handshake and then issues its request unthrottled. So the setting bounded handshake concurrency, which is never the bottleneck — not outstanding calls.

Reproduced with an in-process axum JSON-RPC node (300 ms latency, atomic peak tracking), `max_concurrent_requests: 1`, `max_idle_per_host: 32`, and an 8-way `read_slot` burst behind a barrier:

```
observed peak concurrent in-flight server-side requests: 8
```

**Peak 8, cap 1.**

The fan-out this was meant to bound is real, not hypothetical: `discovery-core`'s `sync/incoming_state.rs` and `outgoing_state.rs` drive nested `FuturesUnordered` across up to `max_channels=256` × `max_subchannels=64` per request, each leaf reaching `read_slot`/`read_slots`, multiplied again across concurrent HTTP clients by axum. Under load that is hundreds of uncapped concurrent calls at the upstream node.

## How

The limiter goes on the transport rather than at each call site. `JsonRpcTransport` has exactly two methods — `send_request` and `send_requests` — and every one of `Provider`'s ~30 methods funnels through one of them. A `LimitedTransport` decorator holding a `Semaphore` therefore covers **all** call sites by construction.

The alternative was a semaphore on `RpcBackendInner` acquired at each of the six provider-call sites. That is correct for the file as it stands, but a seventh call site added later would silently bypass a hand-placed `acquire()`. The decorator cannot be bypassed.

A permit spans exactly one HTTP round trip, so no permit is held across a pagination loop or a multi-chunk batch: `read_slots` calls `batch_requests` once per `max_batch_size` chunk and `get_events` once per continuation page, each acquiring and releasing independently. This is what makes a small limit safe under wide fan-out.

Also:
- `RpcBackendError::InvalidMaxConcurrentRequests` — `max_concurrent_requests: 0` was accepted by config deserialization and leaves the backend with no permits, unable to issue any call. No live config path sets it (default `10`, `config.example.toml` `10`, and `apply_env_overrides` does not cover numeric fields), so this is operator-error-only — but nothing rejected it.
- `LimitedTransportError` wraps `HttpTransportError` plus a `LimiterClosed` variant, so permit acquisition does not need an `.expect()`.
- `tower`'s feature moves `"limit"` → `"util"`. `tower` stays a dependency (`api/mod.rs` uses `tower::Layer`, two test modules use `ServiceExt`); `"limit"` was its only remaining use here. `Cargo.lock` drops `tokio-util` accordingly.

`reqwest-middleware` was considered and ruled out: `HttpTransport::new_with_client` takes a concrete `reqwest::Client`, not a generic, so a `ClientWithMiddleware` cannot be substituted without forking `HttpTransport`.

## Tests

New `tests/test_rpc_concurrency_limit.rs` — axum mock node with artificial latency and in-flight/peak counters, echoing each requested storage key back so responses can be matched to requests:

- single-call burst (cap 2, burst 8) — peak within cap
- batch burst (cap 2, `max_batch_size` 2, 6×4 slots → 12 batch calls) — covers `send_requests`, not just `send_request`
- mixed burst under a single permit — asserts peak `== 1` exactly, and that all callers still complete with correct values (so the limiter cannot regress into a deadlock or dropped request)
- zero-limit rejection

Verified the tests have teeth. Reverting to the old behavior (pass-through transport + `connector_layer` restored), same test file unchanged:

```
test test_single_call_burst_respects_request_limit ... FAILED
    node saw 8 concurrent calls, limit is 2
test test_batch_burst_respects_request_limit ... FAILED
    node saw 6 concurrent batches, limit is 2
test test_mixed_burst_completes_under_single_permit ... FAILED
    assertion `left == right` failed: a single permit must serialize the node's view of the burst
      left: 7,  right: 1
test result: FAILED. 1 passed; 3 failed
```

Peaks of **8, 6 and 7** against caps of 2, 2 and 1 — the connector layer bounded nothing, and peak equalled full burst size every time.

## Verification

```
cargo fmt --check          exit 0
cargo clippy --all-targets 0 warnings
cargo test -p discovery-service --all-targets   78 passed, 0 failed
```

Toolchain note: the system `rustc` (1.75) is too old for several pinned dependencies; verified with 1.89.0.

## Specs

Updated `07-rpc-batching.md` (where the bound is enforced, and its unit), `18-configuration.md` (new "Request limit validation" section — semantics plus startup failure on `0`), and `19-scaling-and-capacity.md` (the knob counts one permit per round trip, batches included).

## Cross-layer

No SDK or e2e change needed. Nothing in `sdk/` references `max_concurrent_requests`, `max_idle_per_host`, or the `[rpc]` config shape. `e2e/scripts/README.md:174` mentions `max_concurrent_requests` as a tuning knob in prose only, and remains accurate.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/938)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3173,7 +3173,6 @@ dependencies = [
  "pin-project-lite",
  "sync_wrapper",
  "tokio",
- "tokio-util",
  "tower-layer",
  "tower-service",
  "tracing",
--- a/crates/discovery-service/Cargo.toml
+++ b/crates/discovery-service/Cargo.toml
@@ -34,7 +34,7 @@ rustls = { version = "0.23", default-features = false, features = ["ring", "std"
 rustls-pemfile = "2"
 tls-listener = { version = "0.11", features = ["rustls-ring", "axum"] }
 tokio-rustls = "0.26"
-tower = { version = "0.5", features = ["limit"] }
+tower = { version = "0.5", features = ["util"] }
 tower-http = { version = "0.6", features = ["cors", "timeout"] }
 
 # Serialization
--- a/crates/discovery-service/specs/07-rpc-batching.md
+++ b/crates/discovery-service/specs/07-rpc-batching.md
@@ -5,6 +5,6 @@ RPC calls can and should be batched. Some reads require sequential probing, but 
 - Parallelize across channels for the same recipient.
 - Parallelize across subchannels within multiple channels.
 - Batch nullifier existence checks for many derived nullifiers at once.
-- Use bounded concurrency to avoid overload of the RPC backend.
+- Use bounded concurrency to avoid overload of the RPC backend. `rpc.max_concurrent_requests` is enforced inside the JSON-RPC transport, so every call reaching the node — single-method or batched — is counted regardless of which code path issued it. A permit covers exactly one HTTP round trip, so paginated `starknet_getEvents` scans and chunked batch reads release their permit between round trips instead of holding one for the whole operation.
 
 Batching will not eliminate probing, but it reduces round trips and improves throughput.
--- a/crates/discovery-service/specs/18-configuration.md
+++ b/crates/discovery-service/specs/18-configuration.md
@@ -63,6 +63,8 @@ server_budget = 10000
 max_request_body_bytes = 102400
 ```
 
+**Request limit validation:** `max_concurrent_requests` bounds how many JSON-RPC calls the RPC backend keeps in flight against the node; each single-method call and each batch request counts as one. It must be at least 1 — startup fails with a configuration error when it is `0`, since zero permits admit no outbound calls at all.
+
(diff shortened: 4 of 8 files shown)


Yoni-Starkware (approved):
:lgtm:

> @Yoni-Starkware reviewed 8 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/938#-:-OzktHdk1rzN3s3iEdXj:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on avi-starkware).

PR #406, opened 2026-02-04
refactor: consume entire panic data from execute_and_panic

arad-starkware:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/406)

PR #942, opened 2026-08-11
refactor: rename sub-accounts to shadow accounts

odednaor:
Renames the concept in the Cairo anonymizer, in every TypeScript layer that reads it, and at the wallet seam, so one vocabulary spans contract, SDK, client and e2e.

## What changed

**Cairo** — package `sub_account_anonymizer` → `shadow_account_anonymizer`:

| Before | After |
|---|---|
| `SubAccountAnonymizer` | `ShadowAccountAnonymizer` |
| `ISubAccountAnonymizer` | `IShadowAccountAnonymizer` |
| `SubAccountInfo` | `ShadowAccountInfo` |
| `SubAccountDeployed` | `ShadowAccountDeployed` |
| `get_sub_accounts` | `get_shadow_accounts` |
| `get_sub_account` | `get_shadow_account` |
| `get_sub_account_class_hash` | `get_shadow_account_class_hash` |

**SDK / client** — `SubAccount*` → `ShadowAccount*`, `resolveSubAccounts` → `resolveShadowAccounts`, config key `subAccountAnonymizerAddress` → `shadowAccountAnonymizerAddress`, builder namespace `build().subaccounts(dapp)` → `build().shadowAccounts(dapp)` (also now camelCase).

**Wallet seam**, per [starknet-specs#406](https://github.com/starkware-libs/starknet-specs/pull/406) (merged): `STRK20_SUBACCOUNT_INVOKE_ACTION` → `STRK20_SHADOW_ACCOUNT_INVOKE_ACTION`, `"subaccount_invoke"` → `"shadow_account_invoke"`.

`sdk/src/internal/anonymizer-abi.ts` is regenerated from the new build artifact, not hand-edited.

## Breaking — read before merging

**This requires a contract upgrade.** The renamed views and the event change their selectors and event keys, so this SDK cannot read a pre-upgrade anonymizer instance. Nothing auto-publishes (`SDK Publish Prerelease` is `workflow_dispatch`), so merging is safe on its own, but **do not publish the SDK before mainnet and Sepolia are upgraded**.

Ship order:

1. Merge
2. Declare the new class — its hash differs from the deployed `0x071118cb…`, so Voyager source-verification needs redoing
3. Grant `UPGRADE_GOVERNOR` — currently held by nobody, so `add_new_implementation` would revert (`GOVERNANCE_ADMIN` is held, and is self-administered; upgrade delay is 0, so grant + add + replace fit in one multicall)
4. Upgrade Sepolia, then mainnet
5. Publish the SDK

Two further consequences:

- Historical `ShadowAccountDeployed` events keep the **old** key, so an indexer reading across the upgrade must match both keys permanently.
- The action discriminant is the wire contract with the wallet, so a wallet implementing a spec revision before #406 will not recognize actions built here. The strk20 types remain a local shim because starknet.js has not released them (absent from 10.5.0).

## Deliberately unchanged

- **`privacy_compute` / `privacy_invoke_with_computation`** — the privacy pool dispatches these by hardcoded selector (`utils.cairo:86`, `:89`), so renaming them would force a pool upgrade too. This is also why the rename's blast radius is only the read views: no contract calls the renamed entrypoints.
- **`starkware_accounts::sub_account::SubAccount`**, the per-commitment account class. It lives in the `starkware-starknet-utils` git dependency, and the anonymizer pins its class hash in storage with **no setter** — so a renamed class could never be adopted, and adding a setter to adopt one would shift every *undeployed* shadow-account address (deployed ones are read from storage and safe). Documented at `e2e/src/shadow-account-setup.ts:58` so it doesn't read as a missed rename.

## Verification

| Layer | Result |
|---|---|
| `scarb build`, `scarb fmt --check` | pass |
| Cairo — 4 packages | **350 passed, 0 failed** (3 pre-existing ignored reference-data generators) |
| Rust — `fmt`, `clippy --all-targets` (0 warnings), `test` | **229 passed, 0 failed** |
| SDK — lint + full suite incl. devnet | **262 passed, 0 failed** |
| Client — lint + tests | **66 passed, 0 failed** |
| e2e — devnet suite | **25 passed, 0 failed** |
| e2e — prettier, eslint | pass |

The two renamed e2e devnet suites exercise `get_shadow_accounts` and `get_shadow_account_class_hash` against a live devnet-deployed `ShadowAccountAnonymizer`, so the new selectors are confirmed end-to-end rather than only type-checked.

## Also included

`client/package-lock.json` gets a 2-line refresh: its embedded `file:../sdk` snapshot still recorded `0.14.3-rc.3`, so `npm ci` resolved a stale linked SDK.

## Follow-ups worth their own PRs

Two pre-existing gaps found while getting the suites to run locally — neither is caused by this change:

- `e2e`'s `npm run build:contracts` misses four targets CI builds (privacy release profile, privacy test target, `test-token`, `ekubo`), and the SDK's devnet tests additionally need `scarb build --features test_contracts`. Without them the suites fail on `ENOENT` for a contract artifact.
- Nothing pins `starknet-devnet` in `.tool-versions`, although the harness requires 0.8.0 (`--proof-mode`). The Rust `discovery-service` integration tests also flake under parallel `cargo test` from concurrent devnet spawns — they pass 11/11 serialized or run alone.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/942)


Yoni-Starkware (approved):
:lgtm:

> @Yoni-Starkware partially reviewed 47 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/942#-:-OzkxJwU8Xsqx7Bz2-vG:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on odednaor).

PR #951, opened 2026-08-13
fix(demo): continue history pagination when a page returns no transactions

Yoni-Starkware:
Client-side companion to #866. Without it, that fix is invisible in the demo.

## Why

#866 makes an **empty-but-incomplete** history page a normal response. The withdrawal scan over the gap between note blocks is budget-bounded, so an account idle for a long stretch spends an entire page covering empty range and returns zero transactions with `historyComplete: false` plus an advanced cursor.

The demo could not express that state:

- `HistoryPanel` gated "Load More" on `transactions.length > 0`
- `useHistory` auto-fetches exactly one page

So an empty first page was a dead end: the cursor was never sent back, and the account's history stayed invisible.

Observed on `starknet-privacy-testnet` running the #866 image (`discovery-service:pr-866`), at the default `SERVER_BUDGET=10000`:

```
history request user=0x5499b2… num_subchannels=21 begin_block_number=None
withdrawal_events: chunked window fetched from_block=12015896 to_block=13393467
                   window_bottom=12374588 chunks_granted=995 num_withdrawal_events=0
fetch_transactions done num_transactions=0 history_complete=false
                   budget_remaining=8 begin_block_number=Some(12374587)
```

995 chunks × `COST_EVENTS_CHUNK` (10) + 21 subchannels × `COST_NOTE` (2) = 9992 of 10000. The page reaches back ~1.02M blocks; the account's oldest note sits at 12,015,896, about 359k blocks below where the budget ran out. Every request in the log arrived with `begin_block_number=None` — each was a fresh first page, never a continuation.

## Changes

- Render "Load More" whenever the history is incomplete, empty list included.
- Walk up to `MAX_PAGES_PER_FETCH` pages per fetch, stopping at the first page that carries a transaction or once the history is complete, so a far-behind account converges without one click per empty page. An empty page that does not move `begin_block_number` is surfaced as an error rather than replayed.
- Pass the pinned snapshot block as `blockIdentifier` — the option `fetchHistory` actually accepts. It was spelled `blockRef` and silently dropped, so every page re-snapshotted at a moving chain head. The pin is cleared on failure so a restarted indexer or devnet does not make every later page fail identically.
- Import `IndexerDiscoveryProvider` from the `starknet-sdk` barrel instead of the `@ts-expect-error`-ed deep dist import (which was `any`, exactly why the misnamed option typechecked). The matching `vite.config.ts` alias goes with it; bundle output is unchanged.
- `sdk`: add a request-body test asserting `fetchHistory` sends `blockIdentifier` as `block_ref`, which is the test that would have caught the typo.
- `e2e`: bound the devnet history pagination loop with `MAX_PAGES`, matching its integration sibling.
- `ci`: pin `actions/checkout`, `actions/setup-node`, `dorny/paths-filter` and `setup-universal-sierra-compiler` to full commit SHAs. The org policy now rejects tag references, which is why Client CI and SDK Devnet Test failed at job setup on `main` as well.

That last item surfaced a real latent bug: `usePrivateState` iterated `channelsResult.channels` unguarded, while its other use site 100 lines earlier already guards the same optional field. It would throw `undefined is not iterable` when discovery returns no channels; now guarded with `?? []`.

## Verification

- `npx tsc --noEmit` clean. Confirmed against a stashed baseline: `main` typechecks clean only because the provider is `any` — with it typed, the `usePrivateState` error appears, which is the point.
- `eslint` and `prettier --check` show no new findings. Both already fail on `main` for these files (pre-existing floating-promise errors and formatting drift); I deliberately did not run `prettier --write`, which would have pulled unrelated reformatting into this diff. My added lines are format-clean.
- Raising `SERVER_BUDGET` to 20000 on testnet makes the first page reach past the current gap, which is how #866 was confirmed end-to-end. That is a mitigation, not a substitute: the idle gap grows ~170k blocks/day, so any fixed budget eventually falls short again and the empty-page path returns.

Not covered here: the gap scan still costs ~1000 RPC event queries to discover zero withdrawals. Pre-indexing withdrawal events in the indexer, which already walks every block, would remove the problem class — the `Alternative: pre-index events off-path` note that #866 removed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951)


avi-starkware (commented):
> @avi-starkware+AGNT made 11 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951)* status: 0 of 4 files reviewed, 8 unresolved discussions (waiting on avi-starkware and Yoni-Starkware).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951#-P0aqDf52lIDnzDLVzbk:-P0aqDf52lIDnzDLVzbl:b-dmx5s4) (no related file):*
Two test suggestions that live outside this PR's diff, both optional.

**`sdk/tests/internal/indexer-discovery.test.ts`** is why the `blockRef` typo survived since April: nothing asserts the block pin reaches the wire. `fetchHistory` has zero references anywhere under `sdk/tests`, and the existing `describe("request body validation")` covers `contract_address` and `viewing_key` for `discoverNotes` / `discoverChannels` / `discoverRequirement`, but not `fetchHistory`. A ~10-line `it` asserting the POSTed body has `block_ref === BLOCK_REF` closes it, the file's `mockFetchJson` helper is a drop-in, and that package has vitest installed (4.0.17, suite green at 30 tests). This is the one test I'd actually write.

**`e2e/tests/devnet/history.test.ts:154`** has an unbounded `while (!historyCursor.historyComplete)`, while its sibling at `privacy-starknet-integration.test.ts:349` bounds with `MAX_PAGES`. With #866 making the server error on a no-progress page, bounding this one turns a hung CI job into a red assertion.
___
*[`demo/src/starknet.ts` line 12 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951#-P0apynU7vwR7GoRLr41:-P0apynU7vwR7GoRLr42:barcdc4) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/d9f2e0b9472ae2959354ccfd4e0e26bdcf5a7848/demo/src/starknet.ts#L12)):*
> ```XML
>   type PrivateTransfersInterface,
> } from "starknet-sdk";
> // Type-only, so it costs nothing at runtime but keeps calls through the
> ```

Non-blocking. The barrel already exports this as a value (`sdk/src/index.ts:29`, and `dist/index.d.ts:14` / `dist/index.js:9`), and it's already imported two lines above. So 12-18 can collapse into the `starknet-sdk` block at 7-11, and the `@ts-expect-error` goes with them.

I tested the "avoids pulling in Node-only modules" rationale rather than arguing about it: built both ways against a freshly built `sdk/dist` and got identical output, `index-KyM_bw02.js 701.59 kB | gzip: 196.13 kB`, with `tsc --noEmit` clean either way. The equality is the point, not the hash, since CI emits a different one because it rebuilds the SDK.

One thing to fold in: `demo/vite.config.ts` carries a `resolve.alias` for `starknet-sdk/dist/internal/indexer-discovery.js` that would go dead too.
___
*[`demo/src/hooks/useHistory.ts` at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951#-P0aq9T799RwclXydYiG:-P0aq9T799RwclXydYiH:b-d40z36) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/d9f2e0b9472ae2959354ccfd4e0e26bdcf5a7848/demo/src/hooks/useHistory.ts)):*
File-level note on test coverage, recorded once rather than as a request.

The new page-walking loop has no automated test, and `demo/` has no cheap way to get one: nine scripts in `demo/package.json` and none named `test`, no runner in `demo/node_modules/.bin`, and zero test files referencing `useHistory` or `demo/src`. I'm not asking you to stand up a harness in a bugfix PR.

The counterweight is more interesting. The regression class the `starknet.ts` comment names really is guarded, by the type system, and I mutation-verified it: reverting the option key fails `tsc`. I also confirmed nothing in `e2e/` or `sdk/` encodes stop-on-empty-page, so this PR breaks no existing test. Both repo pagination loops already continue purely on `!historyComplete`.

One thing I could not settle: whether `demo`'s typecheck runs on PRs. No workflow invokes `demo`'s `lint`, and `typescript-ci.yml` is scoped to `sdk`. The Vercel job on this PR does print `Running "npm run build"` -> `tsc -b && vite build`, so it does run here, but that job is skipped for fork PRs.
___
*[`demo/src/hooks/useHistory.ts` line 273 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951#-P0aq83sC7DLjeReIMTf:-P0aq83sC7DLjeReIMTg:b1etmsm) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/d9f2e0b9472ae2959354ccfd4e0e26bdcf5a7848/demo/src/hooks/useHistory.ts#L273)):*
> ```XML
> }
> 
> /** Transactions to collect before handing control back to the reader. */
> ```

Nit. "the reader" isn't used anywhere else in `demo/src`, and this codebase has a distinct viewing-key / view-only concept (`session.ts`, `useAccounts.ts`) that it could be read as. "the caller" or "the UI"?
___
*[`demo/src/hooks/useHistory.ts` line 276 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951#-P0aq1fSCr4LR_w61zr7:-P0aq1fT4m4eGdn7c0Lf:b-scewcd) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/d9f2e0b9472ae2959354ccfd4e0e26bdcf5a7848/demo/src/hooks/useHistory.ts#L276)):*
> ```XML
> const TRANSACTIONS_PER_FETCH = 5;
> 
> /**
> ```

Question about merge order.

This docstring's premise doesn't hold on this PR's base. There, a budget-exhausted gap scan returns `InsufficientBudget` which maps to HTTP 500 (`discovery-service/src/api/types.rs:119-128`), so `fetchHistory` throws at `:333`, the loop dies on page 1, and the cursor never advances. The 200-with-zero-transactions-and-advanced-cursor shape arrives with #866.

I swept for another route to that shape with seven probes (budget sweep 10..100, `max_transactions: 0`, page-full-then-empty-second-page, partial-page-then-next-page, two block-0 fixtures) and found it only with a pool note in genesis block 0, which can't occur since the contract has to be deployed first.

That said, this reads as "much less useful until #866 lands" rather than a hard sequencing requirement. Budget-truncated *partial* pages (`Ok`, non-empty, `history_complete=false`) are reachable today, so the loop does change base behaviour, and dropping the `transactions.length > 0` guard in `HistoryPanel` gives a retry affordance after a 500 that the base UI didn't offer. Just worth confirming the intended order.
___
*[`demo/src/hooks/useHistory.ts` line 283 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951#-P0aqCPO3YyMgQ3ZrV0R:-P0aqCPO3YyMgQ3ZrV0S:b6e1r6r) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/d9f2e0b9472ae2959354ccfd4e0e26bdcf5a7848/demo/src/hooks/useHistory.ts#L283)):*
> ```XML
>  * cursor. Stopping after one page strands such an account on an empty list.
>  */
> const MAX_PAGES_PER_FETCH = 5;
> ```

Informational, no change requested. One click can now issue up to 5 `fetchHistory` calls plus up to 5 `resolveTimestamps` block-fetch batches, where the base issued one. Bounded here, and inherently sequential since each page's cursor feeds the next, so there's nothing to parallelize. Flagging only because #866 is reducing per-request work server-side while this moves work to the client, bounded 5x.

Checked and clear, so you don't have to: `resolveTimestamps` already dedupes block numbers via a `Set` and batches with `Promise.all` (no N+1), `buildDisplayMaps` is correctly hoisted above the loop, and the `provider` added to the `fetchMore` dependency array is a `useMemo(..., [])`-stable reference (`App.tsx:101`), so it introduces no re-render hazard and in fact fixes a stale closure that existed at the merge-base.
___
*[`demo/src/hooks/useHistory.ts` line 331 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951#-P0aq3FxAQYpkqk9Zvfm:-P0aq3FxAQYpkqk9Zvfn:b9tyf86) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/d9f2e0b9472ae2959354ccfd4e0e26bdcf5a7848/demo/src/hooks/useHistory.ts#L331)):*
> ```XML
>         buildDisplayMaps(account, allAccounts, config);
> 
>       let numCollectedTransactions = 0;
> ```

Question. Is collecting to a total intended, rather than stopping at the first productive page?

The stated bug is "a page returns no transactions", for which the minimum fix is breaking on the first productive page. This keeps fetching until it has `TRANSACTIONS_PER_FETCH` in total, so a click landing on a partially productive page (1-4 txs) issues further sequential round trips where the base did one. The accumulator-free variant, `if (newTransactions.length > 0) { setTransactions(...); break; }`, typechecks clean.

Related: `TRANSACTIONS_PER_FETCH` is doing double duty as the per-page `maxTransactions` (`:338`) and the per-click total (`:367`). They happen to be the same number, which is what makes the second break hard to read. A second named constant if collect-to-total is deliberate.

Not asking about the value 5 separately, since it's derivable from `server_budget = 10_000` with `COST_EVENTS_CHUNK = 10` per `EVENTS_COST_CHUNK_SIZE = 1024` blocks, and `MAX_PAGES = 5` is already the convention at `e2e/tests/integration/privacy-starknet-integration.test.ts:346`.
___
*[`demo/src/hooks/useHistory.ts` line 332 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951#-P0aq7KC6HKyKqVi3uCB:-P0aq7KC6HKyKqVi3uCC:b4jn6ql) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/d9f2e0b9472ae2959354ccfd4e0e26bdcf5a7848/demo/src/hooks/useHistory.ts#L332)):*
> ```XML
> 
>       let numCollectedTransactions = 0;
>       for (let pageNumber = 0; pageNumber < MAX_PAGES_PER_FETCH; pageNumber += 1) {
> ```

Question, and easy to decline. This page-walk now exists in three places: here, `e2e/tests/devnet/history.test.ts:153-164`, and `e2e/tests/integration/privacy-starknet-integration.test.ts:349-360`.

Blame says the demo copy is the one that drifted: it passed the non-existent `blockRef` from `33064cde1` (2026-04-05) while both e2e copies passed `blockIdentifier` (`5cf23fbc9`). That's exactly the bug this PR fixes, which is a decent argument for the protocol living in one place.

An `async *fetchHistoryPages` next to `fetchHistory` (`sdk/src/internal/indexer-discovery.ts:388`) would do it, but that's net-new code rather than reuse, and the type fix here already closes the hole that let the drift go unnoticed. Happy for this to be a no.
___
*[`demo/src/hooks/useHistory.ts` line 340 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951#-P0apvcK3vxceqhXAdJf:-P0apvcL0L8Q76wcQnRT:bln2095) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/d9f2e0b9472ae2959354ccfd4e0e26bdcf5a7848/demo/src/hooks/useHistory.ts#L340)):*
> ```XML
>             maxTransactions: TRANSACTIONS_PER_FETCH,
>             historyCursor: historyCursorRef.current,
>             blockIdentifier: blockRefValue.current,
> ```

Non-blocking, but the one I'd actually act on.

Now that the option name is right, the pin takes effect, and it also persists. `blockRefValue.current` is cleared only in the account/pool reset effect, and `validate_block_ref` does no canonicality check on an explicit `block_ref`, so after an indexer or devnet restart every later Load More fails identically. Before this PR each request silently re-resolved to head, because the misnamed key was dropped.

Worth saying the rename is load-bearing, not cosmetic: reverting `blockIdentifier:` to `blockRef:` under the newly typed provider fails `tsc --noEmit` (exit 2) and `tsc -b` (exit 1) with `TS2353: 'blockRef' does not exist in type`. So the demo had been paginating unpinned since `33064cde1`.

Clearing `blockRefValue.current` in the `catch` restores the old recovery without giving up the pin on the happy path.
___
*[`demo/src/hooks/useHistory.ts` line 348 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951#-P0apzhz4SFxUfUaEc6w:-P0apzhz4SFxUfUaEc6x:b-sq85f4) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/d9f2e0b9472ae2959354ccfd4e0e26bdcf5a7848/demo/src/hooks/useHistory.ts#L348)):*
> ```XML
>         setHistoryComplete(page.cursor.historyComplete);
> 
>         let newTransactions = toDisplayTransactions(
> ```

Non-blocking. Same four maps and the same 7-argument `toDisplayTransactions` call as `:399-408`, including `config.paymasterForwarderAddress ? BigInt(...) : undefined`, which the new loop now re-derives once per page even though it's invariant across pages.

Folding the maps and the two config-derived args into `(rawTransactions, account, config, displayMaps)` on the existing `toDisplayTransactions` (`:211`), reusing `buildDisplayMaps` (`:165`) via `type DisplayMaps = ReturnType<typeof buildDisplayMaps>`, typechecks clean. Not quoting a line saving, since mine came out formatting-dependent.
___
*[`demo/src/hooks/useHistory.ts` line 361 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951#-P0aq4MV8iCn3bYNZ4rO:-P0aq4MV8iCn3bYNZ4rP:b-yw9qqj) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/d9f2e0b9472ae2959354ccfd4e0e26bdcf5a7848/demo/src/hooks/useHistory.ts#L361)):*
> ```XML
>         }
> 
>         if (newTransactions.length > 0) {
> ```

Question. `:344` assigns the returned cursor with no equality check, and this appends with no dedupe, unlike `refreshLatest` which filters by `transactionHash` at `:412-413`. A service handing back the same page (unchanged cursor, non-empty transactions) would get those rows appended once per iteration, so a user could see a withdrawal duplicated from one click. Bounded by `MAX_PAGES_PER_FETCH`, so no hang and no unbounded growth.

Being straight about reachability: I couldn't construct a backend path that returns a non-advancing cursor *with* transactions, so triggering this needs a service violating its own reference contract.

Break when `page.cursor` equals the cursor just sent, or lean on #866's server-side guard? Either is a fine answer, I'd just like to know which one is intended.

Yoni-Starkware:
Pushed 7db108e7 addressing the review. Per discussion:

- **`starknet.ts:12` barrel import** — done. `IndexerDiscoveryProvider` now comes from `starknet-sdk`, the deep dist import, its `@ts-expect-error` and the `vite.config.ts` alias are gone. `tsc --noEmit` clean, `vite build` succeeds.
- **`useHistory.ts:273` "the reader"** — reworded; the constant is now `TRANSACTIONS_PER_PAGE` with "Transactions requested per page."
- **`useHistory.ts:276` merge order** — intended to land after #866; that is the PR this is a companion to, and the docstring describes the post-#866 contract. It still changes base behaviour on budget-truncated partial pages and adds the retry affordance, so nothing here depends on #866 to be correct, only to be useful.
- **`useHistory.ts:331` collect-to-total** — took the minimum fix. The walk stops at the first page carrying any transaction, so the accumulator and the double-duty constant are gone.
- **`useHistory.ts:340` persisted pin** — `blockRefValue.current` is cleared in the `catch`, so the next attempt re-snapshots at head.
- **`useHistory.ts:348` `toDisplayTransactions`** — folded into `(rawTransactions, account, config, displayMaps)` with a named `DisplayMaps` type; both call sites use it.
- **`useHistory.ts:361` non-advancing cursor** — both. Breaking at the first non-empty page means a repeated page can no longer duplicate rows within one click. For the empty-page case, the loop throws when `begin_block_number` did not move, which surfaces as the panel error instead of replaying up to 5 identical requests. #866's server-side guard remains the primary defence.
- **`useHistory.ts:332` shared page-walk helper** — declining for now, as offered; the barrel import closes the drift hole.
- **Tests** — added the `fetchHistory` request-body test asserting `block_ref === BLOCK_REF` (`sdk` suite now 31 tests, green), and bounded `e2e/tests/devnet/history.test.ts` with `MAX_PAGES = 5` plus an assertion that the history completed.
- **CI** — Client Lint, Client Build and SDK Devnet Test were failing at job setup because the org now requires actions pinned to full commit SHAs. Pinned them using the SHAs already in use elsewhere in the repo. `main` has the same failure.

Yoni-Starkware (commented):
> @Yoni-Starkware reviewed 1 file and resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951)* status: 0 of 10 files reviewed, 1 unresolved discussion.

avi-starkware (approved):
> @avi-starkware+AGNT reviewed 10 files and all commit messages, made 3 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951)* status: all files reviewed, 1 unresolved discussion (waiting on Yoni-Starkware).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951#-P0aqDf52lIDnzDLVzbk:-P1ARUsM7wW36gf3cDKz:b-blp5je) (no related file):*
<details><summary><i>Previously, avi-starkware+AGNT (Agent AGNT for Avi Cohen) wrote…</i></summary><blockquote>

Two test suggestions that live outside this PR's diff, both optional.

**`sdk/tests/internal/indexer-discovery.test.ts`** is why the `blockRef` typo survived since April: nothing asserts the block pin reaches the wire. `fetchHistory` has zero references anywhere under `sdk/tests`, and the existing `describe("request body validation")` covers `contract_address` and `viewing_key` for `discoverNotes` / `discoverChannels` / `discoverRequirement`, but not `fetchHistory`. A ~10-line `it` asserting the POSTed body has `block_ref === BLOCK_REF` closes it, the file's `mockFetchJson` helper is a drop-in, and that package has vitest installed (4.0.17, suite green at 30 tests). This is the one test I'd actually write.

**`e2e/tests/devnet/history.test.ts:154`** has an unbounded `while (!historyCursor.historyComplete)`, while its sibling at `privacy-starknet-integration.test.ts:349` bounds with `MAX_PAGES`. With #866 making the server error on a no-progress page, bounding this one turns a hung CI job into a red assertion.
</blockquote></details>

Both landed in 42bc24b1. The SDK test asserts `block_ref` on the wire and the e2e loop is bounded with a `historyComplete` assertion after it. Resolving.
___
*[`demo/src/hooks/useHistory.ts` line 361 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951#-P0aq4MV8iCn3bYNZ4rO:-P1ARVYs-5T6sNiHJn8L:bic79i5) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/d9f2e0b9472ae2959354ccfd4e0e26bdcf5a7848/demo/src/hooks/useHistory.ts#L361)):*
<details><summary><i>Previously, avi-starkware+AGNT (Agent AGNT for Avi Cohen) wrote…</i></summary><blockquote>

Question. `:344` assigns the returned cursor with no equality check, and this appends with no dedupe, unlike `refreshLatest` which filters by `transactionHash` at `:412-413`. A service handing back the same page (unchanged cursor, non-empty transactions) would get those rows appended once per iteration, so a user could see a withdrawal duplicated from one click. Bounded by `MAX_PAGES_PER_FETCH`, so no hang and no unbounded growth.

Being straight about reachability: I couldn't construct a backend path that returns a non-advancing cursor *with* transactions, so triggering this needs a service violating its own reference contract.

Break when `page.cursor` equals the cursor just sent, or lean on #866's server-side guard? Either is a fine answer, I'd just like to know which one is intended.
</blockquote></details>

Satisfied. For the record: the new guard covers the empty-page case, and breaking on the first productive page means a non-empty stalled page can't re-append within one click. Across two clicks it still could, but that needs a service returning transactions without advancing the cursor, which I couldn't produce against the reference backend. Fine to leave as is.
___
*[`-- commits` at r3](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951#-P1ARbfg0KQ-441E7HvO:-P1ARbfg0KQ-441E7HvP:b-symiuu):*
Non-blocking, one-line amend. The last bullet of `42bc24b1` says it pins `actions/checkout`, `setup-node`, `paths-filter` and the sierra compiler to SHAs and that this fixed Client CI and SDK Devnet Test. That was true of r2, which touched three workflow files. The rebase onto `ce65fd6a` (#989, which had already pinned them on `main`) reverted those changes, so r3 touches no `.github/` files and those jobs went green because of the rebase, not this commit.

Worth dropping the bullet so `git log` doesn't send someone looking for workflow changes that aren't here.

avi-starkware (approved):
:lgtm:

> @avi-starkware made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951#-:-P1AS-q2BYR2yU6TCnSm:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on Yoni-Starkware).

avi-starkware (approved):
> @avi-starkware reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951)* status: all files reviewed, 1 unresolved discussion (waiting on Yoni-Starkware).

avi-starkware (approved):
> @avi-starkware reviewed 10 files.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/951)* status: all files reviewed, 1 unresolved discussion (waiting on Yoni-Starkware).

PR #955, opened 2026-08-20
feat(proof-interceptor): read open-note screening policies from the pool

avi-starkware:
The interceptor needs a target's open-note screening policy to know whether to
screen it, and reads it from the pool over RPC rather than from config, so a
governance change needs no paired deploy. An unresolvable read is null, never a
guessed policy: the caller must fail closed, since an assumed Exempt lets a
deposit through unscreened and an assumed Required mints an attestation the pool
rejects. The default timeout matches this service's elliptic-proxy budget, the
other outbound call between a prove request and its answer; discovery-service's
30s RpcConfig::request_timeout is the closer protocol match but belongs to a
background indexer. check_screening_policies.py exists because no CI job
regenerates the pool ABI from Cairo, so nothing else connects the policy list
Cairo owns to the union this decode narrows to.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/955)

PR #954, opened 2026-08-20
feat(proof-interceptor): derive the shadow account address of a prove request

avi-starkware:
A prove request's ComputeAndInvoke on the anonymizer determines the shadow
account its open notes fund, so the interceptor derives that address locally —
the SDK's shared formulas over the request's own felts (user, viewing key, dapp
name, nonce) — rather than reading the chain. The viewing key is a secret: it
never leaves the derivation and an unparseable one costs only it, never the
screening of the rest of the transaction.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/954)

PR #953, opened 2026-08-20
feat(sdk): derive a shadow account's address off-chain

avi-starkware:
The shadow account derivation moves from a builder method into pure functions
(partial commitment, commitment, address) so callers holding raw felts — the
proof interceptor now, the SDK's mock prover next — share one formula with the
builder. A committed cross-language vector pins all three felts against the
anonymizer's own entry points, and CI asserts PRIMER_CLASS_HASH agrees between
Cairo and TS.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/953)

PR #950, opened 2026-08-12
feat(anonymizer): a shadow account is its interaction's associated address

avi-starkware:
The anonymizer answers the pool's delegated screening query: `get_screening_subjects` names the
shadow account the deposits passed through, which is the address compliance must cover — the dapp
pays the shadow account, and the anonymizer collects from there into an open note. Its parameters
mirror `privacy_invoke_with_computation`, because the pool forwards that call's calldata verbatim,
so the view's parameter list is the deserializer for it; `calls` is declared and never read, since
which shadow account acts decides the address, not what it does. No open notes means no answer.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/950)

PR #957, opened 2026-08-23
feat(proof-interceptor): screen the address an open-note deposit requires

avi-starkware:
Each address is gated on the policy that makes the pool ask for it, because an
attestation the pool has no subject for reverts with UNEXPECTED_SCREENING: a
deposit screens its own depositor, a Required open-note depositor screens
itself, and a Delegated one screens the shadow account its compute-invoke runs
through. Deriving that address is per-contract work implemented for the
anonymizer alone, so any other delegated depositor is refused rather than left
to revert unattested, as is a policy the pool cannot be read for.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/957)

### DIFF
--- a/proof-interceptor/src/screened-address.ts
+++ b/proof-interceptor/src/screened-address.ts
@@ -1,0 +1,159 @@
+// src/screened-address.ts
+import { CairoCustomEnum, num } from "starknet";
+import { decodeClientActions, normalizeFelt } from "./pool-transaction.js";
+import type { PoolCallActions } from "./pool-transaction.js";
+import type { ProveTxnV3 } from "./types.js";
+import { getShadowAccountAddress } from "./shadow-account.js";
+import type { OpenNoteScreeningPolicyClient } from "./screening-policy.js";
+
+type PolicyReader = Pick<OpenNoteScreeningPolicyClient, "getPolicy">;
+
+export interface ScreenedAddressConfig {
+  poolAddress: string;
+  anonymizerAddress: string;
+}
+
+type InvokeAction = "InvokeExternal" | "ComputeAndInvoke";
+
+/**
+ * The contract whose invoke funds a transaction's open notes — the one the pool holds an open-note
+ * policy for — and the action it is driven through.
+ */
+interface OpenNoteDepositor {
+  address: string;
+  action: InvokeAction;
+}
+
+/** The pool takes one attestation per transaction, so a delegated depositor puts up at most one address. */
+type DelegatedAddress = Extract<
+  ScreenedAddress,
+  {
+    kind: "one" | "none" | "unknownDelegate" | "undeterminedShadowAccount";
+  }
+>;
+
+export type ScreenedAddress =
+  | { kind: "none" }
+  | { kind: "one"; address: string }
+  | { kind: "conflict" }
+  | { kind: "unreadablePolicy" }
+  | { kind: "unknownDelegate" }
+  | { kind: "undeterminedShadowAccount" };
+
+/**
+ * The address a prove request must be screened for.
+ *
+ * Screening an address the pool did not ask for is not a harmless extra: it rejects an attestation
+ * it has no subject for with `UNEXPECTED_SCREENING`.
+ */
+export async function getScreenedAddress(
+  transaction: ProveTxnV3,
+  config: ScreenedAddressConfig,
+  policyReader: PolicyReader
+): Promise<ScreenedAddress> {
+  // Calldata the pool cannot parse reverts on its own, so nothing in it needs screening.
+  const poolCall = decodeClientActions(transaction, config.poolAddress);
+  if (poolCall === null) return { kind: "none" };
+
+  const addresses = new Set<string>();
+
+  // A deposit is screened on its own depositor. The screening policy list does not apply to a
+  // `TransferFrom` the user signs for themselves.
+  if (poolCall.actions.some((action) => action.activeVariant() === "Deposit")) {
+    addresses.add(poolCall.userAddress);
+  }
+
+  const openNoteDepositor = getOpenNoteDepositor(poolCall.actions);
+  if (openNoteDepositor !== null) {
+    const policy = await policyReader.getPolicy(openNoteDepositor.address);
+    switch (policy) {
+      case "Exempt":
+        break;
+
+      case "Required":
+        addresses.add(openNoteDepositor.address);
+        break;
+
+      case "Delegated": {
+        const delegated = getDelegatedAddress(
+          poolCall,
+          openNoteDepositor,
+          config
+        );
+        if (delegated.kind === "one") addresses.add(delegated.address);
+        else if (delegated.kind !== "none") return delegated;
+        break;
+      }
+
+      case null:
+        return { kind: "unreadablePolicy" };
+
+      default:
+        return unscreenableUnderUnhandledPolicy(policy);
+    }
+  }
+
+  if (addresses.size > 1) return { kind: "conflict" };
+  const [address] = addresses;
+  return address === undefined ? { kind: "none" } : { kind: "one", address };
+}
+
+/** The address a delegated open-note depositor puts up for the deposits its invoke funds. */
+function getDelegatedAddress(
+  poolCall: PoolCallActions,
+  openNoteDepositor: OpenNoteDepositor,
+  { anonymizerAddress }: ScreenedAddressConfig
+): DelegatedAddress {
+  // A plain invoke is exempt under `Delegated`; only a compute-invoke puts up an address.
+  if (openNoteDepositor.action !== "ComputeAndInvoke") return { kind: "none" };
+
+  if (openNoteDepositor.address !== normalizeFelt(anonymizerAddress)) {
+    console.error(JSON.stringify({ error: "unknown_delegated_depositor" }));
+    return { kind: "unknownDelegate" };
+  }
+
+  const shadowAccount = getShadowAccountAddress(poolCall, anonymizerAddress);
+  if (shadowAccount === null) {
+    console.error(JSON.stringify({ error: "shadow_account_undetermined" }));
+    return { kind: "undeterminedShadowAccount" };
+  }
+  return { kind: "one", address: shadowAccount };
+}
+
+/**
+ * Under the invariant that an open note must be funded within the transaction that creates it, any
+ * transaction carrying a `CreateOpenNote` action has an open-note depositor.
+ */
+function getOpenNoteDepositor(
+  actions: CairoCustomEnum[]
+): OpenNoteDepositor | null {
+  if (!actions.some((action) => action.activeVariant() === "CreateOpenNote")) {
+    return null;
+  }
+  for (const action of actions) {
+    const variant = action.activeVariant();
+    if (variant !== "InvokeExternal" && variant !== "ComputeAndInvoke")
+      continue;
+    const { contract_address } = action.unwrap() as {
+      contract_address: bigint;
+    };
+    return {
+      address: normalizeFelt(num.toHex(contract_address)),
+      action: variant,
+    };
+  }
+  return null;
+}
+
+/**
+ * Refuses a policy the switch above does not handle, and fails closed if one reaches it at runtime.
+ * The `never` parameter is the point: `scripts/check_screening_policies.py` keeps
+ * {@link OpenNoteScreeningPolicy} tracking the Cairo enum, so a new variant fails to compile here
+ * instead of falling through and screening nobody.
+ */
+function unscreenableUnderUnhandledPolicy(policy: never): ScreenedAddress {
+  console.error(
+    JSON.stringify({ error: "screening_policy_unavailable", policy })
+  );
+  return { kind: "unreadablePolicy" };
+}
--- a/proof-interceptor/src/screening-interceptor.ts
+++ b/proof-interceptor/src/screening-interceptor.ts
@@ -5,7 +5,10 @@ import type {
   TransactionInterceptor,
   Verdict,
 } from "./interceptor.js";
-import { decodeClientActions, isSinglePoolCall } from "./pool-transaction.js";
+import { isSinglePoolCall } from "./pool-transaction.js";
+import { OpenNoteScreeningPolicyClient } from "./screening-policy.js";
+import { getScreenedAddress } from "./screened-address.js";
+import type { ScreenedAddress } from "./screened-address.js";
 import type { ProveTxnV3 } from "./types.js";
 import {
   screeningResults,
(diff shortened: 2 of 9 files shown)


avi-starkware:
* **#964** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/964?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#963** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/963?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#962** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/962?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#958** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/958?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#957** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/957?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/957?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#955** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/955?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#954** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/954?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#953** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/953?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#950** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/950?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#949** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/949?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#948** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/948?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#947** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/947?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#946** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/946?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#945** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/945?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#944** <a href="https://app.graphite.com/github/pr/starkware-libs/starknet-privacy/944?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

remollemo (commented):
> @remollemo reviewed 5 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/957)* status: 5 of 9 files reviewed, 2 unresolved discussions (waiting on avi-starkware, ofir-starkware, and Yoni-Starkware).

___
*[`proof-interceptor/src/screened-address.ts` line 95 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/957#-P-nIQQ35xMb-liSMRDb:-P-nIQQ35xMb-liSMRDc:b-yroz4h) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/cbd6319b9a0a666944700fcaa7d21bae46fbb412/proof-interceptor/src/screened-address.ts#L95)):*
> ```XML
> 
>       default:
>         return { kind: "unimplementedDelegate" };
> ```

why the default is `unimplDelegate` ? looks like an outdated relict



_Code quote:_
```XML
  return { kind: "unimplementedDelegate" };
```
___
*[`proof-interceptor/src/screened-address.ts` line 104 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/957#-P-nIx756nmiaYwYAyYr:-P-nIx756nmiaYwYAyYs:b-2azyn3) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/cbd6319b9a0a666944700fcaa7d21bae46fbb412/proof-interceptor/src/screened-address.ts#L104)):*
> ```XML
> }
> 
> /** The address a delegated open-note depositor puts up for the deposits its invoke funds. */
> ```

claude's comments are worse than bad



_Code quote:_
```XML
/** The address a delegated open-note depositor puts up for the deposits its invoke funds. */
```
___
*[`proof-interceptor/src/screened-address.ts` line 119 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/957#-P-nJcmz3hPdacVvThER:-P-nJcmz3hPdacVvThES:b6faq2e) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/cbd6319b9a0a666944700fcaa7d21bae46fbb412/proof-interceptor/src/screened-address.ts#L119)):*
> ```XML
>     );
>     return { kind: "unimplementedDelegate" };
>   }
> ```

the error flow and message here is misleading.

IIUC - you stipulate that the depositor is the anon_address, ok,
but what does it mean `delegated_screening_unimplemented` ??

why not simply some kind of mismatch error?



_Code quote:_
```XML
  if (openNoteDepositor.address !== normalizeFelt(anonymizerAddress)) {
    console.error(
      JSON.stringify({ error: "delegated_screening_unimplemented" })
    );
    return { kind: "unimplementedDelegate" };
  }
```

remollemo (approved):
> @remollemo reviewed 4 files.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/957)* status: all files reviewed, 2 unresolved discussions (waiting on avi-starkware, ofir-starkware, and Yoni-Starkware).

proof-interceptor/src/screening-interceptor.ts
```diff
@@ -26,6 +28,12 @@ export interface ScreeningConfig {
   maxRetries: number;
   totalTimeoutMs: number;
   poolAddress: string;
+  /** A Starknet JSON-RPC endpoint, used to read open-note screening policies from the pool. */
```
ofir-starkware: both comments look redundant

proof-interceptor/src/screening-interceptor.ts
```diff
@@ -86,18 +81,31 @@ export class ScreeningInterceptor implements TransactionInterceptor {
...
       transaction,
-      this.config.poolAddress
+      this.config,
+      this.policies
     );
-    if (addresses.length === 0) return { action: "allow" };
+    if (screened.kind === "none") return { action: "allow" };
+    // Reasons are opaque codes. They surface to the client as JSON-RPC error `data`, so they must
+    // not reveal an address.
+    if (screened.kind === "conflict") {
```
ofir-starkware: Might be cleaner to do switch cases, but very optional and arguable, your call if you want to do it, resolve if you want

proof-interceptor/src/screened-address.ts
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * The address a prove request must be screened for.
+ *
+ * Screening an address the pool did not ask for is not a harmless extra: it rejects an attestation
+ * it has no subject for with `UNEXPECTED_SCREENING`.
+ */
+export async function getScreenedAddress(
+  transaction: ProveTxnV3,
+  config: ScreenedAddressConfig,
+  policies: PolicyReader
```
ofir-starkware:
```suggestion
  policyReader: PolicyReader
```

ofir-starkware (changes requested):
Only nitpicks in my comments here, but there are a few package scope issues that needs to be fixed right away in my opinion:


before this PR merges:
1. tests directory does not get type checks, so you might get false positives and not catch regressions - include /tests in your tsconfig. I'm pretty sure even tests in this PR would fail type checks

And in general:
1. You should enforce runtime validations of your service configurations and of http requests. Typescript enforces type safety in compilation, but these come in runtime. Other packages in the repo uses zod FYI
2. the line `transaction: transaction as unknown as ProveTxnV3` in `rpc.ts` can lead to runtime failures. It's called type casting, and you basically tell the TS compiler here to ignore the actual underlying object and its structure. We can talk about this more in depth if you want, but you can tell your claude to not do typecasting as long as possible, and in here specifically to validate all fields, including calldata. for example, if the rpc would return a calldata list of `[1,2,3,4,5,6,7]` your service will panic will likely crash the process.
3. A few redundant dependencies - `@google-cloud/functions-framework`, `vite`
4. I would pin down dependency versions, and move starknet.js to the latest major or at least to be consistent across the repo
5. im pretty sure the docker image is overloaded with dev dependencies, tests etc.

proof-interceptor/src/screened-address.ts
avi-starkware: Applied, in #963 rather than here — that PR fixes the type-check gap you found, and the rename rides along with it.

proof-interceptor/src/screening-interceptor.ts
avi-starkware: Dropped the `rpcUrl` one. Kept a one-line doc on `anonymizerAddress`: the name doesn't say that its interactions are what make a shadow account screenable.

proof-interceptor/src/screening-interceptor.ts
avi-starkware: Kept it as a `Record` keyed by `ScreenedAddress["kind"]` — a new kind is then a compile error rather than a missing case that silently screens. Say the word if you'd still rather see a `switch`.

avi-starkware (commented):
> @avi-starkware+AGNT made 6 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/957)* status: 3 of 9 files reviewed, 3 unresolved discussions (waiting on ofir-starkware, remollemo, and Yoni-Starkware).

___
*[`proof-interceptor/src/screened-address.ts` line 95 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/957#-P-nIQQ35xMb-liSMRDb:-P-pQLvQ9lMeeA7h3eOl:bhhnpan) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/cbd6319b9a0a666944700fcaa7d21bae46fbb412/proof-interceptor/src/screened-address.ts#L95)):*
<details><summary><i>Previously, remollemo wrote…</i></summary><blockquote>

why the default is `unimplDelegate` ? looks like an outdated relict
</blockquote></details>

Relict indeed — it's gone. The default arm is now `unscreenableUnderUnhandledPolicy(policy: never)`, so a new Cairo policy variant fails the build instead of falling through to screening nobody.
___
*[`proof-interceptor/src/screened-address.ts` line 104 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/957#-P-nIx756nmiaYwYAyYr:-P-pQMcY6lWD_WDY-MXi:bnt121b) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/cbd6319b9a0a666944700fcaa7d21bae46fbb412/proof-interceptor/src/screened-address.ts#L104)):*
<details><summary><i>Previously, remollemo wrote…</i></summary><blockquote>

claude's comments are worse than bad
</blockquote></details>

Swept the file: the two-sentence `Delegated` comment is one line, the `never`-parameter docstring is shorter, and the type docstrings are trimmed. This one-liner stays — it names which address is derived, which the signature doesn't.
___
*[`proof-interceptor/src/screened-address.ts` line 119 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/957#-P-nJcmz3hPdacVvThER:-P-pQNP72O1czEDbwsWu:bytqtfc) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/cbd6319b9a0a666944700fcaa7d21bae46fbb412/proof-interceptor/src/screened-address.ts#L119)):*
<details><summary><i>Previously, remollemo wrote…</i></summary><blockquote>

the error flow and message here is misleading.

IIUC - you stipulate that the depositor is the anon_address, ok,
but what does it mean `delegated_screening_unimplemented` ??

why not simply some kind of mismatch error?
</blockquote></details>

Renamed to `unknown_delegated_depositor`, internal kind `unknownDelegate`; the SDK's copies of the string moved with it in #962.

Not a mismatch error because the anonymizer isn't the only legal delegate — the pool can list any contract `Delegated`. The condition is that this interceptor can't derive an address for *that* depositor, and the seam stays open for a second one.

avi-starkware (commented):
All five are done, as PRs. **#963** type-checks `tests/` — four files didn't compile, `e2e.test.ts` among them, whose `ScreeningConfig` had gone stale unnoticed.

**#965** puts zod on the RPC envelope and removes the cast: `ProveTxnV3` is inferred from the schema, so `calldata: [1,2,3]` is refused rather than trusted; **#964** does the same for the service config. **#966** pins versions and drops `vite` — `@google-cloud/functions-framework` stays, `tests/e2e.test.ts` hosts elliptic-proxy through it. **#967** moves to starknet.js 10.5.0, matching the SDK. **#968** takes the image 394 → 322 MB; the SDK's `node_modules` was never pruned.

#965-#968 are a separate stack off `main`, so they don't wait on this one.

> @avi-starkware+AGNT made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/957#-:-P-pQbyA6CId3sBRIroV:b-af4e4d)* status: 3 of 9 files reviewed, 3 unresolved discussions (waiting on ofir-starkware, remollemo, and Yoni-Starkware).

proof-interceptor/src/screening-interceptor.ts
ofir-starkware: No need, resolved

ofir-starkware (approved):

ofir-starkware (approved):
:lgtm:

> @ofir-starkware made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/957#-:-P-sv9jn8cQeGXWewkKJ:bnfp4nl)* status: 3 of 9 files reviewed, 3 unresolved discussions (waiting on avi-starkware, remollemo, and Yoni-Starkware).

proof-interceptor/src/screened-address.ts
avi-starkware: Done.

proof-interceptor/src/screening-interceptor.ts
avi-starkware: Done.

proof-interceptor/src/screening-interceptor.ts
avi-starkware: Done.

avi-starkware (commented):
> @avi-starkware made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/957)* status: 3 of 9 files reviewed, 3 unresolved discussions (waiting on ofir-starkware, remollemo, and Yoni-Starkware).

PR #970, opened 2026-08-25
fix(discovery-service): return typed errors, and pin the Rust toolchain

avi-starkware:
`Format & Clippy` is failing on pull requests that contain no Rust — 953 among them. Clippy 1.91 added `result_unit_err`, and two public functions in `discovery-service` return `Result<(), ()>`.

**The lints.** `Indexer::run` and `Shutdown::run` both discarded the error they were holding. The indexer already had an `IndexerError` in hand and threw it away one line before returning; it now returns it. The shutdown listener gets a `ShutdownError` naming its two real failures: a signal handler that could not be registered, and a notification with no subscriber left to receive it. Both now match `ApiServer::run`, which has returned its own error type all along.

Callers are unchanged — `main.rs` routes all three through `flatten()`, which is generic over the error type and erases it.

**The pin.** These lints arrived with a toolchain release, not with a change: the workflows use `dtolnay/rust-toolchain@stable`, so a new clippy turns whichever pull request is open at the time red. `rust-toolchain.toml` pins 1.91.0 with `rustfmt` and `clippy`, which rustup reads in CI and locally. Bumping it becomes a deliberate commit that can carry its own lint fixes, instead of an unrelated branch absorbing them.

**Verified** on 1.91.0: `cargo fmt --check` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo test --workspace --lib --bins` 212 passed. The two new `Shutdown::notify` tests were mutation-proven — replacing the error mapping with `.or(Ok(()))` fails `notify_reports_that_nobody_is_listening` and only that one. Devnet-backed integration tests were not run locally; CI covers them.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/970)

### DIFF
--- a/crates/discovery-service/src/indexer.rs
+++ b/crates/discovery-service/src/indexer.rs
@@ -84,7 +84,7 @@ impl<C: ChainState> Indexer<C> {
     ///
     /// Only recoverable errors (connection, subscription, receive) trigger a retry.
     /// Non-recoverable errors (unsubscribe, close) cause immediate failure.
-    pub async fn run(&mut self) -> Result<(), ()> {
+    pub async fn run(&mut self) -> Result<(), IndexerError> {
         info!("Indexer started");
 
         loop {
@@ -108,7 +108,7 @@ impl<C: ChainState> Indexer<C> {
                 }
                 Err(e) => {
                     error!("Indexer error (non-recoverable): {}", e);
-                    return Err(());
+                    return Err(e);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


remollemo (approved):
:lgtm:

> @remollemo reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/970#-:-P-sAwoB7erb7cN9r20p:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on avi-starkware).

PR #991, opened 2026-09-10
fix(discovery-service): reject zero-size history pages, and correct the cursor docs

Yoni-Starkware:
Extracted from #866 so that changes unrelated to that PR's gap-scan bug fix can land on their own. #866 is now stacked on top of this branch.

Each commit is independent of the follow-up fix and of the other two.

## 1. `fix(discovery-service): reject max_transactions == 0 with 400`

`validate_bound` only rejects values *above* the maximum, so a zero page size reached the scan. There it breaks out of the loop before any work, and can never satisfy the completion condition (`result.len() < max_transactions` is false at `0`), so every page returns an empty `200` with an unchanged cursor. A client paginating until `history_complete` — which the spec instructs, and which both e2e loops do — never terminates.

Verified empirically rather than by reading: a probe calling `fetch_transactions` with `max_transactions = 0` and a deliberately huge budget returns `Ok(empty)` with `history_complete: false` and an untouched cursor on every attempt.

No caller can regress: the field is required (no serde default), the SDK defaults to 50, and the demo/e2e use 1, 5 and 10. A client sending `0` today is already stuck in that loop.

## 2. `docs(discovery-service): the first history request omits begin_block_number`

`begin_block_number` is an `Option<u64>` and the SDK sends it as `undefined`, but the spec told clients to set it to `0` on the first request. Following that literally pins the scan's upper bound at block 0: every note is skipped as above the bound, the page comes back empty and marked complete, and the cursor is burned so a retry cannot recover.

Also corrects "resolves from chain head" — the code resolves from the pinned snapshot block (`block_ref`, or the current head when omitted), which for `Tag(PreConfirmed)` is `cached_head + 1`.

## 3. `test(discovery-core): name the one-block budget shared by history tests`

Three history tests each recomputed the same budget expression inline. Bound once as `ONE_BLOCK_BUDGET` so the arithmetic lives in one place and the cost constants can change without touching each test. Test-only; no production code.

## Testing

`cargo fmt --check` clean · `cargo clippy -p discovery-core -p discovery-service --all-targets` 0 warnings · `cargo test` 156 (discovery-core) + 58 (discovery-service) + all integration suites pass.

The new validator test was proven to fail when its guard is disabled (`if max_transactions == 0` → `if false` ⇒ `test_history_cursor_zero_max_transactions_rejected` fails on its own `unwrap_err`), per the repo's "prove a new test can fail" rule. A boundary test pins that `max_transactions == 1` is still accepted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/991)


avi-starkware (approved):
:lgtm:

> @avi-starkware reviewed 3 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/991#-:-P1TjaB39MrP0sIndTRI:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on Yoni-Starkware).

___
*[`crates/discovery-service/src/api/validators.rs` line 175 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/991#-P1Tiayw6OwAqAU_t2Se:-P1Tiayw6OwAqAU_t2Sf:b-cd1khb) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/d95639df6c66278ff47df2edeb9a4ee2d1a00b71/crates/discovery-service/src/api/validators.rs#L175)):*
> ```XML
> ///
> /// A zero-transaction page can never advance the scan or mark it complete, so a
> /// client paginating until `history_complete` would loop forever.
> ```

I think it can be removed (describes a state that will no longer exist after this change).
___
*[`crates/discovery-service/src/api/validators.rs` line 546 at r1](https://reviewable.io/reviews/starkware-libs/starknet-privacy/991#-P1TioJYEvkbMl9TmoRs:-P1TioJYEvkbMl9TmoRt:b-2vdnhx) ([raw file](https://github.com/starkware-libs/starknet-privacy/blob/d95639df6c66278ff47df2edeb9a4ee2d1a00b71/crates/discovery-service/src/api/validators.rs#L546)):*
> ```XML
>         let limits = ValidationLimits::default();
>         validate_history_cursor(&HistoryCursor::default(), 1, &limits).unwrap();
>     }
> ```

Consider unifying to one test with two cases.



_Code quote:_
```XML
    #[test]
    fn test_history_cursor_zero_max_transactions_rejected() {
        let limits = ValidationLimits::default();
        let cursor = HistoryCursor::default();

        let (status, error) = validate_history_cursor(&cursor, 0, &limits).unwrap_err();
        assert_eq!(status, StatusCode::BAD_REQUEST);
        assert_eq!(error.error.code, error_codes::INVALID_REQUEST);
        assert!(error.error.message.contains("max_transactions"));
    }

    #[test]
    fn test_history_cursor_min_max_transactions_accepted() {
        let limits = ValidationLimits::default();
        validate_history_cursor(&HistoryCursor::default(), 1, &limits).unwrap();
    }
```

Yoni-Starkware (commented):
> @Yoni-Starkware resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/991)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on avi-starkware).

Yoni-Starkware (commented):
> @Yoni-Starkware resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/starknet-privacy/991)* status: 2 of 3 files reviewed, all discussions resolved (waiting on avi-starkware).

